PR #80, opened 2020-02-24
Cleanup for Article Series Component

jestapinski:
This PR cleans up the series component, specifically adding the following things:

- Fixes text overflow ellipsis
- Fixes border/bullet interactions
- Cleans up mobile styling
- Moves series logic in `article.js` into own component.

Since bullets are text, I used a filled in bullet behind to hide the border going through the bullet.

Screen Shot 2020-02-24 at 3 48 04 PM
Screen Shot 2020-02-24 at 3 48 21 PM

### DIFF
--- a/src/components/dev-hub/series.js
+++ b/src/components/dev-hub/series.js
@@ -55,6 +55,9 @@ const Breadcrumb = styled('li')`
 const DescriptiveText = styled(P)`
     color: ${colorMap.greyLightThree};
     font-size: ${fontSize.tiny};
+    @media ${screenSize.upToMedium} {
+        padding-bottom: ${size.tiny};
+    }
 `;
 
 const SeriesBreadcrumbs = styled('div')`
--- a/src/components/dev-hub/series.js
+++ b/src/components/dev-hub/series.js
@@ -6,18 +6,37 @@ import { H3, P } from './text';
 import {
     colorMap,
     fontSize,
-    gradientMap,
     layer,
+    lineHeight,
     screenSize,
     size,
 } from './theme';
 
-const BULLET_BOX_WIDTH = '14px';
-const BULLET_SIZE = '22px';
+const BULLET_BOX_WIDTH = '30px';
+const BULLET_SIZE = 36;
+
+// Needed to allow more magenta to account for proper text coloring
+const BULLET_GRADIENT = `linear-gradient(
+    315deg,
+    ${colorMap.sherbet} 0%,
+    ${colorMap.salmon} 40%,
+    ${colorMap.magenta} 100%
+);`;
+
+const BORDER_GRADIENT = `linear-gradient(
+    0deg,
+    ${colorMap.sherbet} 0%,
+    ${colorMap.salmon} 49.99%,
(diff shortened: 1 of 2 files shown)


src/templates/article.js
```diff
@@ -60,18 +60,37 @@ const getContent = nodes => {
...
+    slugTitleMapping,
+}) => {
     try {
-        const relevantSeries = allSeries[seriesName];
-        const mappedSeries = relevantSeries.map(s => ({
-            slug: s,
-            title: slugTitleMapping[s][0].value,
-        }));
-        return mappedSeries;
+        if (!!currentArticleSeries) {
```
greggb: The try/catch and if blocks seem different from how we write all our other components. Is there a structural reason for this new pattern?

src/components/dev-hub/series.js
```diff
@@ -6,67 +6,50 @@ import { H3, P } from './text';
...
     size,
 } from './theme';
 
-const BULLET_BOX_WIDTH = '14px';
-const BULLET_SIZE = '22px';
+const BULLET_BOX_WIDTH = '30px';
+const BULLET_SIZE = 36;
+
+// Needed to allow more magenta to account for proper text coloring
+const BULLET_GRADIENT = `linear-gradient(
```
jbouzi12: Now that we have a lot of gradients everywhere, maybe we should create a helper function so that they can be created on the fly when we need them?

jbouzi12: This may or may not be a post-launch cleanup item though. What do you guys think?

melissarudofsky: can we add these to the gradientMap in the theme file?

src/templates/article.js
```diff
@@ -66,31 +66,23 @@ const ArticleSeries = ({
...
-    }
-    return null;
+    // Handle if this article is not in a series or no series are defined
+    if (!allSeries || !currentArticleSeries) return null;
+    const relevantSeries = allSeries[currentArticleSeries];
+    // Handle if this series is not defined in the top-level content TOML file
+    if (!relevantSeries || !relevantSeries.length) return null;
+    const mappedSeries = relevantSeries.map(s => ({
+        slug: s,
+        title: slugTitleMapping[s][0].value,
```
greggb: Do we need to guard this, or is the mapped value guaranteed to have a first item with a value by this point?

jestapinski: We can guard. It should but one can never be too safe.

src/components/dev-hub/series.js
jestapinski: We can, but I think they are only used in one place so that is why I didn't want to move them into theme. We can definitely parameterize though and I can work on that after this PR.

jestapinski:
> Just had two observations and otherwise looks great!
> 
> * the `more from series` text seems to have a bit more top/bottom padding than the mocks
> * the series title (ex: `Storybook Series`) has a bit more bottom padding on mobile than the mocks.

Nice catches! I'll adjust them a bit.

jestapinski:
Thanks!

PR #84, opened 2020-02-26
Codeblock Button and Language Changes

jestapinski:
This PR does two things for the codeblock:

- Writers are trying to put in language options that don't match with the leafygreen codeblock languages (i.e., writers put in `csharp` when leafygreen wants `csp`). Sue is going to determine a mapping from writer language to leafygreen language and then we can add that in. For now, setting the language to `auto` resolves this.
- Adds a prop to the `Button` to disable the arrow

### DIFF
--- a/src/components/dev-hub/button.js
+++ b/src/components/dev-hub/button.js
@@ -179,9 +179,9 @@ const getArrow = ({ pagination, primary, secondary }) => {
  * @property {string?} props.to
  */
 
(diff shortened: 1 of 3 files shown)


jestapinski:
Thanks! Going to wait and see if the writers want to use the leafygreen languages before merging.

jestapinski:
It looks like the writers want to use leafygreen's language specs, so I removed that piece from this PR. Thanks!

PR #87, opened 2020-02-26
Related Internal Articles

jestapinski:
This PR is not to be merged yet, it is a WIP for how I want to tackle related articles.

We get `meta.related` back, which (once the parser is updated), will return an array of nodes (not strings as it is with some test data). These nodes could be links to other articles or external based on the wiki.

We don't get the images for internal articles back with this node, so this WIP suggests adding it in `createPages` if it exists, and using a placeholder otherwise (still need to move the placeholder into createPages). This way we can safely add the image to the related node and reference it every time.

Curious if this approach is best? I left samples of the internal reference doc nodes in `article.js`. I don't think Role Docs would be useful here?

Screen Shot 2020-02-26 at 11 17 07 AM

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -1,4 +1,5 @@
 const path = require('path');
+const dlv = require('dlv');
 const fs = require('fs').promises;
 const mkdirp = require('mkdirp');
 const { Stitch, AnonymousCredential } = require('mongodb-stitch-server-sdk');
@@ -84,6 +85,18 @@ const constructDbFilter = () => ({
     commit_hash: process.env.COMMIT_HASH || { $exists: false },
 });
 
+const getRelatedPagesWithImages = pageNodes => {
+    const related = dlv(pageNodes, 'query_fields.related', []);
+    let relatedPageInfo = [];
+    if (related.length) {
+        relatedPageInfo = related.map(r => ({
+            image: dlv(RESOLVED_REF_DOC_MAPPING, [r.target, 'atf-image'], null),
+            ...r,
+        }));
+    }
+    return relatedPageInfo;
+};
+
 exports.sourceNodes = async () => {
     // setup env variables
     const envResults = validateEnvVariables();
@@ -148,6 +161,10 @@ exports.createPages = async ({ actions }) => {
                 getNestedValue(['ast', 'options', 'template'], pageNodes)
             );
             const slug = getPageSlug(page);
+            if (pageNodes.query_fields) {
+                const relatedPages = getRelatedPagesWithImages(pageNodes);
+                pageNodes['query_fields'].related = relatedPages;
+            }
             createPage({
                 path: slug,
                 component: path.resolve(`./src/templates/${template}.js`),
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -87,17 +87,14 @@ const constructDbFilter = () => ({
 
 const getRelatedPagesWithImages = pageNodes => {
     const related = dlv(pageNodes, 'query_fields.related', []);
-    let relatedPageInfo = [];
-    if (related.length) {
-        relatedPageInfo = related.map(r => ({
-            image: dlv(
-                RESOLVED_REF_DOC_MAPPING,
-                [r.target, 'query_fields', 'atf-image'],
-                null
-            ),
-            ...r,
-        }));
-    }
+    const relatedPageInfo = related.map(r => ({
+        image: dlv(
+            RESOLVED_REF_DOC_MAPPING,
+            [r.target, 'query_fields', 'atf-image'],
+            null
+        ),
+        ...r,
+    }));
     return relatedPageInfo;
 };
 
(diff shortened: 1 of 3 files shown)


gatsby-node.js
```diff
@@ -84,6 +85,22 @@ const constructDbFilter = () => ({
     commit_hash: process.env.COMMIT_HASH || { $exists: false },
 });
 
+const getRelatedPagesWithImages = pageNodes => {
+    const related = dlv(pageNodes, 'query_fields.related', []);
+    let relatedPageInfo = [];
```
greggb: nit: since we're guaranteeing an array for `related` we can just return the below map without checking for length since the map will return and empty array when run on an empty array  - e.g. `return related.map(<...>)`

This assumes that `'`query_fields.related` would be undefined and not null.

src/templates/article.js
```diff
@@ -74,6 +76,48 @@ const getSeries = (allSeries, slugTitleMapping, seriesName) => {
...
+    padding: 70px 120px;
+    margin: 0 auto;
+    height: 575px;
+`;
+
+const RelatedCards = styled('div')`
+    display: flex;
+`;
+
+const RelatedArticles = ({ related, slugTitleMapping }) => {
```
greggb: Can you pull this out into a separate file?

src/templates/article.js
```diff
@@ -74,6 +76,48 @@ const getSeries = (allSeries, slugTitleMapping, seriesName) => {
...
+            <H4>Related</H4>
+            <RelatedCards>
+                {related.map(r => {
+                    const target = r.target;
+                    const slug = r.target && r.target.slice(1, r.target.length);
+                    const image = r.image || ARTICLE_PLACEHOLDER;
+                    const title = dlv(
+                        slugTitleMapping,
+                        [slug, 0, 'value'],
+                        'Title Not Found'
```
greggb: should this be `''` so we don't output an error message just in case? We can log something instead maybe?

greggb:
Approach makes sense to me. Just had a few nits for the WIP stuff

jestapinski:
The [wiki](https://wiki.corp.mongodb.com/display/DE/devhub-article+Reference) seems to indicate there can be related Links to external sources, but not sure this is worth tackling right now in this PR. This PR can handle references to other internal docs and if that is okay for now it is ready for another look.

If not we can come back to it when we can clarify the structure for external link nodes returned from the parser.

greggb:
Yea, let's go with internal for now so we can continue forward and extend it later on with the logic from `components/Link` to check if external/internal

jestapinski:
This PR is ready for a review. The current related data is just string literals, so the cards don't redirect and no titles should be seen. The final parser will return a node type for each related article.

Screen Shot 2020-02-27 at 10 57 42 AM
Screen Shot 2020-02-27 at 10 57 53 AM

src/components/dev-hub/related-articles.js
```diff
@@ -0,0 +1,78 @@
...
+                    const title = dlv(slugTitleMapping, [slug, 0, 'value'], '');
+                    if (title === '') {
+                        console.error(
+                            `No title found for this article ${slug}`,
+                            slugTitleMapping
+                        );
+                    }
+                    /* TODO: Case on doc to link internal vs external */
+                    return (
+                        <Card
```
melissarudofsky: small thing but we should probably have a key here!

src/components/dev-hub/related-articles.js
```diff
@@ -0,0 +1,78 @@
...
+const RelatedArticles = ({ related, slugTitleMapping }) => {
+    if (!related || !related.length) return null;
+    return (
+        <RelatedContainer>
+            <H4>Related</H4>
+            <RelatedCards>
+                {related.map(r => {
+                    const target = r.target;
+                    const slug = r.target && r.target.slice(1, r.target.length);
+                    const image = r.image || ARTICLE_PLACEHOLDER;
```
melissarudofsky: is this something we should be doing everywhere?

src/components/dev-hub/related-articles.js
```diff
@@ -0,0 +1,78 @@
...
+    padding: 30px;
+    @media ${screenSize.mediumAndUp} {
+        padding: 70px ${size.xxlarge};
+    }
+`;
+
+const RelatedCards = styled('div')`
+    display: flex;
+    @media ${screenSize.upToMedium} {
+        flex-direction: column;
```
melissarudofsky: instead of going directly from a row to a column, should the cards wrap?

jestapinski: Gregg and I talked about this and it is life to just switch to column at medium for now.

src/components/dev-hub/related-articles.js
jestapinski: Not sure?

src/components/dev-hub/related-articles.js
jestapinski: I added a key, but because title might not exist for some reason yet (if they want to use External links, which we still need to support anyway) I tacked on the index as well to ensure uniqueness. Not sure if this is the right thing to do.

src/components/dev-hub/related-articles.js
greggb: image

jestapinski: whoops

jestapinski:
Thanks!

PR #88, opened 2020-02-26
Article Hero Cleanup

jestapinski:
This PR:

- Removes the hardcoding on breadcrumbs and will add another breadcrumb if `meta.type` is supplied, with a link to that page assuming it is going to be `/tag/${meta.type}` for now.
- Removes products and languages from the header tag list and adds `/tag` to the link so they also link to the expected page

### DIFF
--- a/src/components/dev-hub/blog-tag-list.js
+++ b/src/components/dev-hub/blog-tag-list.js
@@ -59,7 +59,7 @@ const BlogTag = ({ children, ...props }) => (
 );
 
 // eslint-disable-next-line no-unused-vars
-const BlogTagList = ({ tags = [], tagLinkBase }) => {
+const BlogTagList = ({ tags = [], tagLinkBase = '/tag' }) => {
     // TODO: add in article link below once finalized
     // const getArticleLink = tagName => `${tagLinkBase}/${tagName}`
     const canExpand = tags.length >= MINIMUM_EXPANDABLE_SIZE;
--- a/src/components/dev-hub/blog-tag-list.js
+++ b/src/components/dev-hub/blog-tag-list.js
@@ -58,31 +58,27 @@ const BlogTag = ({ children, ...props }) => (
     </TagListItem>
 );
 
-// eslint-disable-next-line no-unused-vars
-const BlogTagList = ({ tags = [], tagLinkBase = '/tag' }) => {
-    // TODO: add in article link below once finalized
-    // const getArticleLink = tagName => `${tagLinkBase}/${tagName}`
+const BlogTagList = ({ tags = [] }) => {
     const canExpand = tags.length >= MINIMUM_EXPANDABLE_SIZE;
     // By default any list of blog tags under the minimum expandable size is already expanded
     const [isExpanded, setIsExpanded] = useState(!canExpand);
     const expandList = useCallback(() => setIsExpanded(true), []);
-    const tagData = getTagData(tags, tagLinkBase);
     if (!tags.length) {
         return null;
     }
     return (
         <TagList>
             {isExpanded &&
-                tagData.map(t => (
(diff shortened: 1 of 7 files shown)


src/components/dev-hub/blog-tag-list.js
```diff
@@ -59,7 +59,7 @@ const BlogTag = ({ children, ...props }) => (
 );
 
 // eslint-disable-next-line no-unused-vars
-const BlogTagList = ({ tags = [], tagLinkBase }) => {
+const BlogTagList = ({ tags = [], tagLinkBase = '/tag' }) => {
```
greggb: I think the intention was to have urls with the specific type of tag like `language/<languageName>`. Is that information we can retrieve for each tag?

src/templates/article.js
```diff
@@ -108,6 +108,16 @@ const Article = props => {
...
     const contentNodes = getContent(childNodes);
     const meta = dlv(__refDocMapping, 'query_fields');
+    const articleBreadcrumbs = [
+        { label: 'Home', target: '/' },
+        { label: 'Learn', target: '/learn' },
+    ];
+    if (meta.type && meta.type.length) {
+        articleBreadcrumbs.push({
+            label: meta.type[0].toUpperCase() + meta.type.substring(1),
+            target: `/tag/${meta.type}`,
```
greggb:
```suggestion
            target: `/type/${meta.type}`,
```

src/templates/article.js
```diff
@@ -108,6 +108,16 @@ const Article = props => {
     const childNodes = dlv(__refDocMapping, 'ast.children', []);
     const contentNodes = getContent(childNodes);
     const meta = dlv(__refDocMapping, 'query_fields');
+    const articleBreadcrumbs = [
+        { label: 'Home', target: '/' },
+        { label: 'Learn', target: '/learn' },
+    ];
+    if (meta.type && meta.type.length) {
+        articleBreadcrumbs.push({
+            label: meta.type[0].toUpperCase() + meta.type.substring(1),
```
greggb: I see the `type` field as a string in the console. Are you seeing something else?

src/components/dev-hub/blog-tag-list.js
jestapinski: Yea let's rework these components a bit to allow on a tag-by-tag basis

src/templates/article.js
jestapinski: It is a string, I was looking to capitalize it here!

greggb: aha!

src/templates/article.js
```diff
@@ -117,18 +131,13 @@ const Article = props => {
         <Layout>
             <BlogPostTitleArea
                 // TODO: Pull real image once available
-                // articleImage={meta['atf-image']}
-                articleImage={ARTICLE_PLACEHOLDER}
+                articleImage={meta['atf-image'] || ARTICLE_PLACEHOLDER}
```
jestapinski: I added this back in as well at this point, since the writers can try to add their own images now and we can make sure this is supported properly.

src/utils/get-tag-links-from-meta.js
```diff
@@ -0,0 +1,20 @@
+export const getTagLinksFromMeta = meta => {
+    const mappedTags = mapTagTypeToUrl(meta.tags, 'tag');
+    const mappedLanguageTags = mapTagTypeToUrl(meta.languages, 'language');
+    const mappedProductTags = mapTagTypeToUrl(meta.products, 'product');
+    const allTags = [
```
jestapinski: I didn't think it was necessary now to worry about uniqueness between all the various types of tags.

jestapinski:
Thanks!

PR #91, opened 2020-02-26
Mr/get articles

melissarudofsky:
Display articles on `/learn` page and filter with dropdowns. 

STILL TODO: add query params to url and add an empty state 

not sure how to check if the images will work on not before going to staging!

### DIFF
--- a/src/components/dev-hub/card-list.js
+++ b/src/components/dev-hub/card-list.js
@@ -3,6 +3,8 @@ import styled from '@emotion/styled';
 import Card from './card';
 import Button from './button';
 import { screenSize, size } from './theme';
+import { withPrefix } from 'gatsby';
+import { getNestedText } from '../../utils/get-nested-text';
 
 const CardContainer = styled('div')`
     align-items: center;
@@ -33,18 +35,21 @@ const CenterBlock = styled('div')`
 export default ({ items = [], limit = 9 }) => {
     const [visibleCards, setVisibleCards] = useState(limit);
     const hasMore = items.length > visibleCards;
-    console.log(items);
     return (
         <>
             <CardContainer>
                 {items.slice(0, visibleCards).map(item => (
                     <ArticleCard
                         key={item['_id']}
-                        image={item['atf-image']}
-                        tags={[...item.languages, ...item.products]}
-                    >
-                        {item['meta-description'][0].children[0].value}
-                    </ArticleCard>
+                        image={withPrefix(item['atf-image'])}
+                        tags={[
+                            ...item.tags,
+                            ...item.languages,
+                            ...item.products,
+                        ]}
+                        title={getNestedText(item['title'])}
+                        description={getNestedText(item['meta-description'])}
+                    />
                 ))}
             </CardContainer>
             {hasMore && (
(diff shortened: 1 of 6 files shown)


src/components/dev-hub/card-list.js
```diff
@@ -38,12 +40,16 @@ export default ({ items = [], limit = 9 }) => {
...
                     <ArticleCard
-                        key={item['meta-description']}
-                        image={item['atf-image']}
-                        tags={item.languages}
-                    >
-                        {item['meta-description']}
-                    </ArticleCard>
+                        key={item['_id']}
+                        image={withPrefix(item['atf-image'])}
+                        tags={[
```
jestapinski: I'm adding some helpers for the tags in my PR [here](https://github.com/mongodb/devhub/pull/88/files#diff-6b2033282b007c4d3f0b56a23c5b8e02), since these fields could be optional!

src/components/dev-hub/card-list.js
```diff
@@ -38,12 +40,16 @@ export default ({ items = [], limit = 9 }) => {
...
                 {items.slice(0, visibleCards).map(item => (
                     <ArticleCard
-                        key={item['meta-description']}
-                        image={item['atf-image']}
-                        tags={item.languages}
-                    >
-                        {item['meta-description']}
-                    </ArticleCard>
+                        key={item['_id']}
+                        image={withPrefix(item['atf-image'])}
```
jestapinski: I can't think of a great way to check images locally, was running into a similar problem with articles.

src/components/dev-hub/filter-bar.js
```diff
@@ -82,6 +84,8 @@ export default ({ heading = 'All Articles' }) => {
                         name="product"
                         choices={products}
                         defaultText="Product"
+                        value={filterValue.products}
```
jestapinski: I noticed when filtering by multiple options (language and product) that the number next to the filtered type didnt change to reflect the other filter being applied. Not sure what the right thing to do here is but it looks great!

greggb:
@melissarudofsky Can we use urls instead of query params?

src/components/dev-hub/card-list.js
greggb: Yes, please add a comment here so we don't forget to update it. If there is an article without a language tag it'll crash the renderer

src/components/dev-hub/filter-bar.js
```diff
@@ -70,7 +70,9 @@ export default ({ heading = 'All Articles' }) => {
             callStitch(metadata, 'products', setProducts);
         }
     }, [metadata, languages.length, products.length]);
-
+    const handleChange = (value, type) => {
+        setFilterValue({ ...filterValue, [type]: value });
```
greggb: You might already have this in mind, but the filter changes should probably do something like `GatsbyRouter(or whatever their API is). routeTo('/langage/nodejs')`

src/components/dev-hub/card-list.js
melissarudofsky: updated with the new method! thanks

src/components/dev-hub/filter-bar.js
melissarudofsky: yeah not sure what to do about that because its coming from the stitch function. Either we can not display the number, calculate it on the frontend, or just leave it (its still sort of helpful?).

jestapinski: I think it is fine to leave what we have here for now and come back to it later if need be, what do you think?

PR #100, opened 2020-02-27
add bottom margin to article images

greggb:


greggb:
> Not sure this is the correct place for article images, did you mean in the `HeroBanner` at the top?

This is for images inside article content

PR #103, opened 2020-02-27
remove log output

greggb:
Getting pretty noisy in the build output


jestapinski:
I don't mind this, but one advantage of noisy logs has been that it helps debug issues the writers are having by looking directly at their pages.

greggb:
> I don't mind this, but one advantage of noisy logs has been that it helps debug issues the writers are having by looking directly at their pages.

Fair point, but I think as we get 15-20 articles coming in we're going to want to keep it down. If they're continuing to run into blockers we can selectively re-enable some logs

PR #104, opened 2020-02-27
use query params for filtering

melissarudofsky:


src/pages/learn.js
```diff
@@ -57,14 +58,17 @@ const callStitch = async (metadata, key, callback) => {
...
 
-export default () => {
+export default ({ location }) => {
     const metadata = useSiteMetadata();
     const [articles, setArticles] = useState([]);
-    const [filterValue, setFilterValue] = useState({});
+    const { search = {} } = location;
+    const [filterValue, setFilterValue] = useState(parseQueryString(search));
     useEffect(() => {
+        window.history.pushState({}, '', buildQueryString(filterValue));
```
melissarudofsky: I spent a while trying to use a gatsby or reach-router function here instead but couldn't get it to work correctly. Let me know if you think this is an issue!

greggb: I saw this example
```
import { navigate } from "gatsby";
    navigate("/app/login")
```

Would that work here?

melissarudofsky: it does work but it changes the url so it brings you to the top of the page when it updates

greggb: Ah, could we (just a thought), add an id to the filter bar and link to that with a `#`?

melissarudofsky:
> This looks great, should this piece include a way to clear filters or is that a TODO item?

yeah! I have another PR open with an 'all' option for each dropdown

src/components/dev-hub/select.js
```diff
@@ -89,7 +88,21 @@ const FormSelect = ({
...
         setShowOptions(!showOptions);
     }, [showOptions]);
-
+    const selectOptions = typeof choices !== 'undefined' ? choices : children;
+    /**
+     * This useEffect should only be called once the component first renders with choices,
+     * this should populate the select item with the default choice if there is one
+     */
+    useEffect(() => {
+        if (selectOptions.length && value && value !== '') {
```
greggb: The `&& value` would already be falsey for empty string.

PR #105, opened 2020-02-27
add all option for filter

melissarudofsky:
This PR adds the 'All' option to the filter dropdowns. For now, I did not put a number next the all the option because in order to get that value i would have to make the call for all articles (regardless of the filter) and then count them.

I personally think it looks okay as is, but If we want a number next to all, I'm thinking we should add that to the stitch function that returns the other values.
Screen Shot 2020-02-27 at 1 46 42 PM

### DIFF
--- a/src/pages/learn.js
+++ b/src/pages/learn.js
@@ -57,17 +57,33 @@ const callStitch = async (metadata, key, callback) => {
     const res = await callStitchFunction('fetchDevhubMetadata', metadata, key);
     callback(parseArticles(res));
 };
-
+// strip out the 'All' param from the url and the stitch function key
+const stripAllParam = filterValue => {
+    Object.keys(filterValue).map(key => {
+        if (filterValue[key] === 'all') {
+            delete filterValue[key];
+        }
+    });
+    return filterValue;
+};
 export default ({ location }) => {
     const metadata = useSiteMetadata();
     const [articles, setArticles] = useState([]);
-    const { search = {} } = location;
+    const { search = '', pathname = '' } = location;
     const [filterValue, setFilterValue] = useState(parseQueryString(search));
+
     useEffect(() => {
-        window.history.pushState({}, '', buildQueryString(filterValue));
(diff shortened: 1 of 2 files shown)


src/pages/learn.js
```diff
@@ -57,17 +57,33 @@ const callStitch = async (metadata, key, callback) => {
     const res = await callStitchFunction('fetchDevhubMetadata', metadata, key);
     callback(parseArticles(res));
 };
-
+// strip out the 'All' param from the url and the stitch function key
+const stripAllParam = filterValue => {
+    Object.keys(filterValue).map(key => {
```
jestapinski: Is a `forEach` more appropriate here?

src/pages/learn.js
```diff
@@ -57,17 +57,33 @@ const callStitch = async (metadata, key, callback) => {
...
+        if (filterValue[key] === 'all') {
+            delete filterValue[key];
+        }
+    });
+    return filterValue;
```
jestapinski: nit: maybe I'm misinterpreting this but it seems we modify `filterValue` in place so the references to it should be updated without the need to return anything, but maybe there is some value in passing another reference?

src/pages/learn.js
```diff
@@ -57,17 +57,33 @@ const callStitch = async (metadata, key, callback) => {
...
-    const { search = {} } = location;
+    const { search = '', pathname = '' } = location;
     const [filterValue, setFilterValue] = useState(parseQueryString(search));
+
     useEffect(() => {
-        window.history.pushState({}, '', buildQueryString(filterValue));
+        const filter = stripAllParam(filterValue);
+        const searchParams = buildQueryString(filter);
+        // if the search params are empty, push the pathname state in order to remove params
+        window.history.pushState(
```
jestapinski: Do we also need to handle filtering if the user hits the back button?

src/pages/learn.js
melissarudofsky: yeah this function can definitely be cleaned up! thanks!

PR #106, opened 2020-02-27
Tag Page Routes

jestapinski:
This PR adds logic to pull and generate pages for tags and tag-like things (authors, languages, products, and types).

The data is passed to the `tag.js` template in `context`.

~Still WIP, trying to clean up unnecessary iteration, but happy for any feedback too.~

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -108,16 +108,19 @@ const getRelatedPagesWithImages = pageNodes => {
 const createTagPageType = async (createPage, pageMetadata, stitchType) => {
     const isAuthor = stitchType === 'author';
     const pageType = STITCH_TYPE_TO_URL_PREFIX[stitchType];
-    const res = await stitchClient.callFunction('getValuesByKey', [
-        metadata,
-        stitchType,
-    ]);
+
+    // Query for all possible values for this type of tag
+    const possibleTagValues = await stitchClient.callFunction(
+        'getValuesByKey',
+        [metadata, stitchType]
+    );
 
     const requests = [];
 
-    await res.forEach(async item => {
+    // For each possible tag value, query the pages that exist for it
+    await possibleTagValues.forEach(async tag => {
         const requestKey = {};
-        requestKey[stitchType] = item._id;
+        requestKey[stitchType] = tag._id;
         requests.push(
             stitchClient.callFunction('fetchDevhubMetadata', [
                 metadata,
@@ -129,7 +132,7 @@ const createTagPageType = async (createPage, pageMetadata, stitchType) => {
     const pageData = await Promise.all(requests);
 
     // Once requests finish, map the item with name (and optional image) to the response's return value
-    const itemsWithPageData = res.map((r, i) => ({
+    const itemsWithPageData = possibleTagValues.map((r, i) => ({
         item: r,
         pages: pageData[i],
     }));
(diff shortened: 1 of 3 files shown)


jestapinski:
This is ready for review!

gatsby-node.js
```diff
@@ -84,6 +92,76 @@ const constructDbFilter = () => ({
...
+        );
+    });
+
+    const pageData = await Promise.all(requests);
+
+    // Once requests finish, map the item with name (and optional image) to the response's return value
+    const itemsWithPageData = res.map((r, i) => ({
+        item: r,
+        pages: pageData[i],
+    }));
```
jestapinski: Not sure if there is a cleaner way to avoid race conditions with network data other than to just wait for requests and map with initial objects.

gatsby-node.js
```diff
@@ -152,14 +230,20 @@ exports.createPages = async ({ actions }) => {
...
+                    metadata: pageMetadata,
                     slug,
                     snootyStitchId: SNOOTY_STITCH_ID,
                     __refDocMapping: pageNodes,
                 },
             });
         }
     });
+
+    await createTagPageType(createPage, pageMetadata, 'author');
```
jbouzi12: You can probably use a `Promise.all` here for these.

gatsby-node.js
```diff
@@ -84,6 +92,76 @@ const constructDbFilter = () => ({
     commit_hash: process.env.COMMIT_HASH || { $exists: false },
 });
 
+const createTagPageType = async (createPage, pageMetadata, stitchType) => {
+    const isAuthor = stitchType === 'author';
+    const pageType = STITCH_TYPE_TO_URL_PREFIX[stitchType];
+    const res = await stitchClient.callFunction('getValuesByKey', [
```
jbouzi12: Looking really good! Can you explain what this stitch call in conjunction with line line `108` is doing? I know you explained some of it in person but now that the code is here things will sink in more for me. I haven't touched any of the stitch stuff yet so curious to know how these are working!

jbouzi12: For example,  I see that we get an array of items for this but was I'm wondering why we iterate through each and make other async calls for each item.

jestapinski: Sure! So there are two stitch functions we can call to get data: `getValuesByKey` and `fetchDevhubMetadata`.

`getValuesByKey` takes one of our tag types as a parameter (such as `language`) and returns back all of the values for that key in stitch (such as [`node`, `java`, etc.]) as an array of objects.

For each one of these objects (let's say they are languages), we then need to get all of the article pages for that type of language. Now that we know all of the languages that exist, we can call the `fetchDevhubMetadata` on each one to return to us the article pages associated with it.

You can think of it like a query for what languages exist and then a map to its pages, does that help?

jestapinski:
Thanks! I'll merge after you finish working on the tag template pages @jbouzi12 , since otherwise creating tag pages will break (since no template exists).

gatsby-node.js
greggb: Not really. I think there are a couple request libraries that might make it easier, but probably not worth it since it's build side requests rather than the client

gatsby-node.js
```diff
@@ -134,7 +215,7 @@ exports.sourceNodes = async () => {
 
 exports.createPages = async ({ actions }) => {
     const { createPage } = actions;
-    const metadata = await stitchClient.callFunction('fetchDocument', [
+    const pageMetadata = await stitchClient.callFunction('fetchDocument', [
```
greggb: Just curious, why change this var name?

gatsby-node.js
```diff
@@ -84,6 +92,79 @@ const constructDbFilter = () => ({
...
+
+    const pageData = await Promise.all(requests);
+
+    // Once requests finish, map the item with name (and optional image) to the response's return value
+    const itemsWithPageData = possibleTagValues.map((r, i) => ({
+        item: r,
+        pages: pageData[i],
+    }));
+
+    const PAGES = [];
```
greggb:
```suggestion
    const pageList = [];
```

greggb: I might even suggest a map right on this variable below rather than a forEach, but it's a nit

gatsby-node.js
jestapinski: In `gatsby_node.js` there is also a global `metadata` var. When playing around I noticed the local scope was overriding this one. In this function I think I can change it back, but would like to distinguish in the new helper.

gatsby-node.js
jbouzi12: Ah I see. Yea this was great, thanks!

jbouzi12:
I'll create a WIP PR for my work for the tag template since both rely on eachother for everything to work.

PR #114, opened 2020-02-27
Article Bottom Tags

jestapinski:
this PR adds the bottom area to articles with just tags for now (need to get clarity on where the share links should go).

Screen Shot 2020-02-27 at 4 27 16 PM


greggb:
Can we discuss the share buttons tomorrow? Want to do it in a light-weight way.

PR #115, opened 2020-02-27
Article Updated Date

jestapinski:
This PR adds support for the "Updated Date" field for an article. I need to double check with Sophie the exact name of this field.

Screen Shot 2020-02-27 at 6 04 26 PM

On extremely small screens, let's stack the text to avoid odd wraps.

Screen Shot 2020-02-27 at 6 04 36 PM

### DIFF
--- a/src/components/dev-hub/blog-post-title-area.js
+++ b/src/components/dev-hub/blog-post-title-area.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import { css } from '@emotion/core';
 import styled from '@emotion/styled';
 import BlogTagList from './blog-tag-list';
 import { H2, P } from './text';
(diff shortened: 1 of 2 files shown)


src/components/dev-hub/blog-post-title-area.js
```diff
@@ -18,26 +19,54 @@ const PostMetaLine = styled('div')`
...
+    }
+`;
+
 const DateText = styled(P)`
-    margin-right: ${size.medium};
+    margin-right: ${size.tiny};
     @media ${screenSize.upToLarge} {
         font-size: ${fontSize.xsmall};
     }
+    ${({ withBullet }) => withBullet && bulletStyling};
```
greggb: Bet we could do this with last-child or the `+` selector, but mostly just a noit

jestapinski: Only issue there is if the updated date is not provided we don't want the bullet on the initial published date. I would like to keep as-is?

jestapinski:
Thanks!

src/components/dev-hub/blog-post-title-area.js
greggb: Sure you can keep, but that can be done with css :)

PR #116, opened 2020-02-28
add default author photo

melissarudofsky:


src/components/dev-hub/byline-block.js
```diff
@@ -66,7 +67,10 @@ const ByLine = styled('div')`
     }
 `;
 
-const BylineBlock = ({ authorName = '', authorImage }) => {
+const BylineBlock = ({
+    authorName = '',
+    authorImage = DEFAULT_AUTHOR_IMAGE,
```
jestapinski: We may also need to update some logic in the `article.js` template too. The `withPrefix` function will give a path value even if the image is undefined!

melissarudofsky: Where on `article.js`? We don't call withPrefix on this image until after we have set the default

melissarudofsky:
thank you!

PR #124, opened 2020-02-28
add homepage background image

melissarudofsky:


src/pages/index.js
```diff
@@ -20,9 +20,13 @@ import greenPatternImage from '../images/1x/pattern-green.png';
 import meetupsImage from '../images/1x/Meetups.png';
 import buildImage from '../images/1x/Build.png';
 import GradientUnderline from '../components/dev-hub/gradient-underline';
+import BACKGROUND_IMAGE from '../images/1x/homepage-background.png';
```
jestapinski: Should we use all caps for these imports as a standard?

greggb: why

greggb:
Can we run all the images through https://tinypng.com/ or a related Command line tool. They seem huge

src/pages/index.js
jestapinski: Just curious if we wanted anything special for images

melissarudofsky: made it lower case

PR #127, opened 2020-02-28
Swap updated and publish dates for articles

jestapinski:
Marketing wants the updated date (if available) to come first. This PR swaps the existing order.

Not sure if they also want "Published" before the published date as well?


jestapinski:
Thanks!

PR #129, opened 2020-02-28
Reference External Articles

jestapinski:
This PR adds logic to support external links on related articles. It [looks like the writers are trying to do so](https://github.com/ljhaywar/devhub-content/blob/node-changestreams/source/quickstart/nodejs-change-streams-triggers.txt) at this point.

The type here is a `Reference`, so I went off the structure from the `Reference` component to get the needed fields.

### DIFF
--- a/src/components/dev-hub/related-articles.js
+++ b/src/components/dev-hub/related-articles.js
@@ -43,14 +43,10 @@ const RelatedCards = styled('div')`
 `;
 
 const getCardParamsFromRelatedType = (relatedArticle, slugTitleMapping) => {
-    const name = relatedArticle.name;
+    const name = relatedArticle.name
+        ? relatedArticle.name
+        : relatedArticle.type;
     switch (name) {
-        case 'reference':
-            return {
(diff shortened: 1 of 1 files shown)


src/components/dev-hub/related-articles.js
```diff
@@ -42,23 +42,55 @@ const RelatedCards = styled('div')`
...
+    console.log(relatedArticle);
+    const name = relatedArticle.name
+        ? relatedArticle.name
+        : relatedArticle.type;
+    switch (name) {
+        case 'doc':
+            const target = relatedArticle.target;
+            const slug = target && target.slice(1, target.length);
+            const image = relatedArticle.image || ARTICLE_PLACEHOLDER;
+            const title = dlv(slugTitleMapping, [slug, 0, 'value'], '');
```
greggb: Not for now, but since we do this exact thing with dlv a bunch we should make a util function for this like `get-slug-title` or something

jestapinski: Ah good point, I see what you mean by that now, thanks!

PR #130, opened 2020-02-28
allow images to float

melissarudofsky:


melissarudofsky:
was trying to address, it looks like scale should be working but this updates for aligning
Screen Shot 2020-02-28 at 2 58 13 PM

greggb:
For some reason scale wasn't working, and I didn't want to debug all that code, so I'm just setting it to the width.

PR #135, opened 2020-02-28
Replace encodeURI and non-alphanum chars with regex check

jestapinski:
This PR fixes encodeURI issues with Gatsby by instead replacing non-alphanumeric characters in a url with dashes, so:

mongo 4.4 --> mongo-4-4
.net --> -net
c# --> c-
etc.

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -7,6 +7,9 @@ const {
     validateEnvVariables,
 } = require('./src/utils/setup/validate-env-variables');
 const { getNestedValue } = require('./src/utils/get-nested-value');
+const {
+    getTagPageUriComponent,
+} = require('./src/utils/get-tag-page-uri-component');
 const { getTemplate } = require('./src/utils/get-template');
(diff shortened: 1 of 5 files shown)


gatsby-node.js
```diff
@@ -143,13 +143,12 @@ const createTagPageType = async (createPage, pageMetadata, stitchType) => {
...
-                          .toLowerCase()
-                          .split(' ')
-                          .join('-')
-                  )
-                : encodeURIComponent(name.toLowerCase());
+                ? name
+                      .toLowerCase()
+                      .split(' ')
+                      .join('-')
+                      .replace(/\W/g, '-')
```
greggb: What about writing a helper function with this regex and we can wrap all the necessary strings in this pr with that? 

Also curious why we do a split and join here when the regex should apply to spaces anyway?

jestapinski: Yeah we can simplify this a little bit here.

jestapinski:
Thanks!

PR #143, opened 2020-03-01
check for streams and videos

greggb:
If the mongodb channel is streaming we'll return that data otherwise we'll check for the last recorded video. That will potentially be used as the Twitch feature on the homepage.

I think removing the error and pending state would be more efficient while we're not using them, but we can comment them out when enable the API calls before launch if needed.

### DIFF
--- a/src/utils/use-twitch-api.js
+++ b/src/utils/use-twitch-api.js
@@ -1,0 +1,53 @@
+import { useState, useEffect, useCallback } from 'react';
+import { get } from './request';
+
+const TWITCH_API_ENDPOINT = 'https://api.twitch.tv/helix/';
+const CLIENT_ID = '041r2glmgub2pt357ss0la44j2sz95';
+const MDB_CHANNEL_ID = '467752938';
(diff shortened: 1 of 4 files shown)


jestapinski:
> If the mongodb channel is streaming we'll return that data otherwise we'll check for the last recorded video. That will potentially be used as the Twitch feature on the homepage.
> 
> I think removing the error and pending state would be more efficient while we're not using them, but we can comment them out when enable the API calls before launch if needed.

Yeah I think we can handle this later.

src/utils/use-twitch-api.js
```diff
@@ -0,0 +1,53 @@
...
+            currentStream = streamResp.data[0];
+            setStream(currentStream);
+        }
+        //get videos
+        if (!currentStream && videoLimit) {
+            const videoResp = await get(video_URL + videoLimit, headers);
+            // return the whole array of videos, but ignore pagination for now
+            setVideos(videoResp.data);
+        }
+        try {
```
jbouzi12: Should we put the `get stream` and `get videos` pieces inside this `try`?

greggb: oops, yes.

PR #147, opened 2020-03-02
Render default image without withPrefix if no image exists

jestapinski:
The default author images were accidentally being fed into `withPrefix`, which prevented them from being used properly (since we don't need a prefix for the default image here).

This PR adds a case to check if we are using the default image before using the `withPrefix` function.

### DIFF
--- a/src/components/dev-hub/byline-block.js
+++ b/src/components/dev-hub/byline-block.js
@@ -68,15 +68,20 @@ const ByLine = styled('div')`
     }
 `;
 
-const BylineBlock = ({
-    authorName = '',
-    authorImage = DEFAULT_AUTHOR_IMAGE,
-}) => {
+const BylineBlock = ({ authorImage, authorName = '' }) => {
+    const hasImage = !!authorImage;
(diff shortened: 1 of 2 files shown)


src/components/dev-hub/byline-block.js
```diff
@@ -68,15 +68,20 @@ const ByLine = styled('div')`
...
+const BylineBlock = ({ authorImage, authorName = '' }) => {
+    const hasImage = !!authorImage;
     const authorLink = `/author/${getTagPageUriComponent(authorName)}`;
     return (
         <ByLine>
             <AuthorImage>
-                <img src={withPrefix(authorImage)} alt={authorName} />
+                <img
+                    src={
+                        hasImage
```
greggb: Super nit: move this logic to replace hasImage

`const authorImageSrc = !!authorImage ? withPrefix(authorImage) : DEFAULT_AUTHOR_IMAGE;`

Currently it feels like we're doing the image check twice

jestapinski:
Thanks!

PR #148, opened 2020-03-02
new tab for external links

greggb:
I had some confusion on the difference between RefRole and Reference

RefRole is used for on page anchors and inter-document linking. Reference is basically any link.

We'll handle all links with an http/https prefix as external and open in a new tab for now. We can be smarter about this in the future

### DIFF
--- a/src/components/RefRole.js
+++ b/src/components/RefRole.js
@@ -7,7 +7,7 @@ const RefRole = ({ nodeData: { children, fileid, target, url }, slug }) => {
     // Render intersphinx target links
     if (url) {
         return (
-            <Link to={url} className="reference external">
+            <Link to={url} className="ref-role-inter">
                 {children.map((node, i) => (
                     <ComponentFactory key={i} nodeData={node} />
                 ))}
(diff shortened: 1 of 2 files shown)


src/components/RefRole.js
```diff
@@ -18,7 +18,7 @@ const RefRole = ({ nodeData: { children, fileid, target, url }, slug }) => {
     // Render internal target links
     const link = fileid === slug ? `#${target}` : `${fileid}#${target}`;
     return (
-        <Link href={link}>
+        <Link href={link} className="ref-role-anchor">
```
jestapinski: Just curious where you pulled this class from? Rendered content?

src/components/Reference.js
```diff
@@ -4,7 +4,7 @@ import Link from './Link';
 import { getNestedValue } from '../utils/get-nested-value';
 
 const Reference = ({ nodeData }) => (
-    <Link className="reference external" to={nodeData.refuri}>
+    <Link className="reference" to={nodeData.refuri} target="_blank">
```
jestapinski: Yeah eventually we can swap these links to ours. Cool test for internal here though:
`const external = /^http(s)?:\/\//.test(to);`

src/components/RefRole.js
greggb: Oh, I meant to call this out. It's helpful to debug the various links if we have readable class names, so I just made this up.

PR #153, opened 2020-03-02
Drop text size for nav on mobile

jestapinski:
The nav on mobile was wrapping onto a newline because the text was 18pt. I dropped it to 14pt and it does not wrap.

This PR:
Screen Shot 2020-03-02 at 4 49 10 PM

Before:
Screen Shot 2020-03-02 at 4 49 26 PM


jestapinski:
Thanks!

PR #154, opened 2020-03-02
Static render learn page content

jestapinski:
This PR adds the articles to the learn page to the page context statically generated. There should be no changes on the content shown, but there should no longer be a "loading" state there.

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -296,3 +296,26 @@ exports.onCreateWebpackConfig = ({ stage, loaders, actions }) => {
         },
     });
 };
+
+const getLearnPageArticles = async () => {
+    const documents = await stitchClient.callFunction('fetchDevhubMetadata', [
+        metadata,
+        {},
+    ]);
+    return documents;
+};
+
+exports.onCreatePage = async ({ page, actions }) => {
+    if (page.path === '/learn/') {
+        const { createPage, deletePage } = actions;
+        const allArticles = await getLearnPageArticles();
+        deletePage(page);
+        createPage({
+            ...page,
+            context: {
+                ...page.context,
+                allArticles,
+            },
+        });
+    }
+};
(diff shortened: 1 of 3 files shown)


gatsby-node.js
```diff
@@ -296,3 +296,26 @@ exports.onCreateWebpackConfig = ({ stage, loaders, actions }) => {
...
+        {},
+    ]);
+    return documents;
+};
+
+exports.onCreatePage = async ({ page, actions }) => {
+    if (page.path === '/learn/') {
+        const { createPage, deletePage } = actions;
+        const allArticles = await getLearnPageArticles();
+        deletePage(page);
```
jestapinski: Weird paradigm for adding page context, but this is what [Gatsby recommends](https://www.gatsbyjs.org/docs/creating-and-modifying-pages/).

src/components/dev-hub/related-articles.js
```diff
@@ -43,7 +43,6 @@ const RelatedCards = styled('div')`
 `;
 
 const getCardParamsFromRelatedType = (relatedArticle, slugTitleMapping) => {
-    console.log(relatedArticle);
```
jestapinski: This was also getting noisy in logs and related articles seem to render properly now, so figured this would be a great time to remove.

melissarudofsky:
Very cool! We will still need to call the stitch function when a user uses the dropdown filters!

jestapinski:
> Very cool! We will still need to call the stitch function when a user uses the dropdown filters!

Oh interesting, I bet we also can now get around that as well. Let me take a look

jestapinski:
This should be ready for another look. I think since we have the articles already we can filter in-memory to avoid some extra network overhead, what do you all think?

greggb:
Would this enable us to do a lookup on the articles to populate the 3 featured items with this data (rather than copying pasting). <- Related question that doesn't effect merging this PR!

jestapinski:
> Would this enable us to do a lookup on the articles to populate the 3 featured items with this data (rather than copying pasting). <- Related question that doesn't effect merging this PR!

Yes! This is for the index page, correct?

If so, we can add similar logic to `onCreatePage` for the index page and populate them. We would just need to figure out which ones we want to do.

The `onCreatePage` method runs for EVERY page (template or not).

jestapinski:
Thanks all!

PR #155, opened 2020-03-03
Jb/twitch video

jbouzi12:
This PR includes:
- Replaces Twitch video placeholder on home page with either the current live stream or latest video
- if clicked will open a modal so users can watch the video

@greggb Should I uncomment the live stream banner or do you still want this to be done closer to launch?

### Thumbnail Inconsistencies
If you notice I have some logic to update the dimensions on any thumbnail we receive. The Twitch docs say that the end of `thumbnail_urls` contains tokens `{width}x{height}` through which you can update the dimensions. **However,** I found the `thumbnail_url` had some extra space around the dimensions = `%{width}x%{height}`. 

The url isn't encoded so decoding it doesn't remove this space. I have some logic to handle this format but not sure if this is something that will consistently happen. Let me know your thoughts!

Mar-03-2020 10-22-28

### DIFF
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -3,12 +3,8 @@ import styled from '@emotion/styled';
 import Card from '../components/dev-hub/card';
 import MediaBlock from '../components/dev-hub/media-block';
 import Layout from '../components/dev-hub/layout';
-<<<<<<< HEAD
-import { H1, H2, P, SubHeader } from '../components/dev-hub/text';
-=======
 import Notification from '../components/dev-hub/notification';
 import { H1, H2, P, SubHeader, H5 } from '../components/dev-hub/text';
->>>>>>> Implement latest twitch video player into home page
 import {
     colorMap,
     gradientMap,
@@ -25,14 +21,9 @@ import buildImage from '../images/1x/Build.png';
 import GradientUnderline from '../components/dev-hub/gradient-underline';
 import homepageBackground from '../images/1x/homepage-background.png';
 import ProjectSignUpForm from '../components/dev-hub/project-sign-up-form';
-<<<<<<< HEAD
-// import Notification from '../components/dev-hub/notification';
-// import useTwitchApi from '../utils/use-twitch-api';
-=======
 import useTwitchApi from '../utils/use-twitch-api';
 import { Modal } from '../components/dev-hub/modal';
 import VideoEmbed from '../components/dev-hub/video-embed';
->>>>>>> Implement latest twitch video player into home page
 
 const MEDIA_WIDTH = '550';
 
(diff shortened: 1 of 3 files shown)


src/pages/index.js
```diff
@@ -90,18 +92,71 @@ const DescriptiveText = styled(P)`
...
+`;
+
+const ThumbnailButton = styled(Button)`
+    left: 40%;
+    position: absolute;
+    top: 35%;
+`;
+
+const Thumbnail = ({ video }) => {
+    let thumbnailUrl = buildImage; //fallback image if there is none
```
jestapinski: I wonder if it is possible to get a more upscaled image? This one looks a bit grainy on large devices

src/pages/storybook.js
```diff
@@ -515,14 +515,6 @@ export default () => (
                     </GradientUnderline>
                 </H2>
             </Row>
-            <SectionHeader>Icons</SectionHeader>
```
jestapinski: Was there a reason we wanted to remove these icons?

src/pages/index.js
```diff
@@ -90,18 +92,71 @@ const DescriptiveText = styled(P)`
...
+    let dimensionsMatcher = '{width}x{height}';
+    if (video.thumbnail_url) {
+        const containsSpace = video.thumbnail_url.match('%{width}x%{height}');
+        dimensionsMatcher = containsSpace
+            ? '%{width}x%{height}'
+            : dimensionsMatcher;
+        thumbnailUrl = video.thumbnail_url.replace(
+            dimensionsMatcher,
+            `${MEDIA_WIDTH}x${MEDIA_WIDTH}`
```
jestapinski: Do we also need to add the `%`s here after replacing if it contains space?

src/pages/index.js
```diff
@@ -161,14 +213,11 @@ export default () => {
...
-                            <Button
-                                secondary
-                                href="https://www.twitch.tv/mongodb"
-                                target="_blank"
-                                rel="noopener noreferrer"
-                            >
-                                Watch
-                            </Button>
+                            {twitchVideo && (
+                                <Button secondary href={twitchVideo.url}>
```
jestapinski: Do we want to open this in a new tab?

src/pages/storybook.js
jbouzi12: There were duplicate `Icons` sections. I removed this extra section a few weeks ago but noticed its still here. Probably got re-added during a rebase.

src/pages/index.js
jbouzi12: Nope! Those `%` cause the images to error out and should not be there. So this logic not only sets the dimensions but removes those `%`'s as well.

jbouzi12: I wonder how these thumbnails get saved, because these thumbnails are being returned with the `%`'s in there.

src/pages/index.js
jbouzi12: Yea I'll look into a better image than this one.

src/pages/index.js
jbouzi12: Also, do you know if we want a `noreferrer noopener` on these kinds of external links?

src/pages/index.js
greggb: Maybe `src/images/2x/Build@2x.png`

src/pages/index.js
greggb: I think we automatically add those `rel`s when adding a `_blank` target.

jbouzi12:
Wait going to choose a different image because this is already being used.

src/pages/index.js
```diff
@@ -18,6 +18,7 @@ import stichImage from '../images/2x/Stitch-triggers@2x.png';
 import greenPatternImage from '../images/1x/pattern-green.png';
 import meetupsImage from '../images/1x/Meetups.png';
 import buildImage from '../images/1x/Build.png';
+import buildImage3x from '../images/3x/Build@3x.png';
```
greggb: can we use 2x? Maybe it wasn't good enough?

src/pages/index.js
```diff
@@ -90,18 +92,71 @@ const DescriptiveText = styled(P)`
...
+            <Modal
+                dialogContainerStyle={{
+                    height: '90%',
+                    padding: `0 ${size.large}`,
+                    width: '90%',
+                }}
+                transparent
+                triggerComponent={<ThumbnailButton play />}
+            >
+                <VideoEmbed
```
greggb: can we pass an autoplay prop since we're clicking on a 'play' button already

src/pages/index.js
```diff
@@ -90,18 +92,71 @@ const DescriptiveText = styled(P)`
...
+        if (stream) return stream;
+        if (videos && videos.length) return videos[0];
+        return null;
+    }, [stream, videos]);
+
     return (
         <Layout>
             <BackgroundImage>
                 {/* 
                 // TODO enable for launch
```
greggb: Sorry, I misunderstood early, you can uncomment this now since we're making the API call anyway

src/pages/index.js
jbouzi12: We do have one, I'll use that.

jbouzi12:
@greggb I added the autoplay options for twitch based on the [docs](https://github.com/CookPete/react-player) (and [here](https://dev.twitch.tv/docs/embed/video-and-clips)). Testing it in chrome and firefox doesn't seem to autoplay the video though. Let me know if you're experiencing the same or have had similar issues with `ReactPlayer` and autoplay.

greggb:
> @greggb I added the autoplay options for twitch based on the [docs](https://github.com/CookPete/react-player) (and [here](https://dev.twitch.tv/docs/embed/video-and-clips)). Testing it in chrome and firefox doesn't seem to autoplay the video though. Let me know if you're experiencing the same or have had similar issues with `ReactPlayer` and autoplay.

Yea, it won't always work, but we can send the signal to try. They have a heuristic for when to actually autoplay to avoid ads spamming videos

PR #157, opened 2020-03-03
meta and titles

greggb:
Updates some meta info and page titles


greggb:
Think I have a bug in here when re-building. WIP

PR #161, opened 2020-03-03
Add Related Article Flex-Wrap

jestapinski:
This PR adds some styling improvements for related articles (specifically padding and handling >4 related articles).

- Added `flex-wrap` to keep width consistent and move more cards to new rows
- Fixed padding/margin to properly have card hover state

Because of flex-wrapping and a fixed width, I decided to just switch to a `column` flex under the large size. Otherwise cards were awkwardly flushed left.

Screen Shot 2020-03-03 at 1 24 31 PM
Screen Shot 2020-03-03 at 1 24 43 PM
Screen Shot 2020-03-03 at 1 24 59 PM


jestapinski:
Thanks!

PR #165, opened 2020-03-03
Jb/update events preview

jbouzi12:
-Updates Events data to use the next 4 real events from the events api
- onError return the sample events data for eventsPreview


src/pages/community/events.js
```diff
@@ -43,8 +43,8 @@ export default () => {
                     {/* TODO: Add FilterBar */}
                     <H3>All Events</H3>
                 </EventsFilter>
-                {/* TODO: remove below when events_api is working */}
-                <EventsList items={[...sampleEvents, ...sampleEvents]} />
+                {/* TODO: update below to not use sample data when events_api is working */}
+                <EventsList items={[...sampleEvents]} />
```
greggb: Does this code and comment still apply? We're doing all the sample stuff in the hook now

jbouzi12: Good point, we can render events straight from the hook now.

PR #167, opened 2020-03-03
add styles to table

melissarudofsky:
didn't have designs for this so let me know if you can think of a way to make this look better! 
Screen Shot 2020-03-03 at 3 02 02 PM


jbouzi12:
Which article can I use to find this table?

melissarudofsky:
> Which article can I use to find this table?

i'll send you the updated env file and go to http://localhost:8000/quickstart/mongodb-node-tutorial#series-versions

PR #183, opened 2020-03-03
Jb/update featured articles

jbouzi12:
This PR updates the Home and Learn page featured articles based on the MVP List tab in this doc https://docs.google.com/spreadsheets/d/1bLXYHdL6P8shgEk2FJDJGTV8S0WL2VG1gnIikjvUSSo/edit#gid=1611784364


src/pages/learn.js
```diff
@@ -49,6 +49,44 @@ const Article = styled('article')`
     padding: ${size.medium};
 `;
 
+const FEATURED_ARTICLE_URLS = [
```
jestapinski: Is this to replace the build-time prop passed in `gatsby-node.js`?

jbouzi12: Wow I did not realize that was there. I can probably just update the article list here rather than doing the new filtering

jestapinski: Sounds good! Yep that is what I would suggest :)

jbouzi12:
Thanks @jestapinski for pointing out the article filtering in `gatsby-node` so we don't have this redundant filtering in the view.

jbouzi12:
> Thanks for the change!
> 
> Some of the slugs on articles don't match locally, but they are valid on staging so these should work once we get an updated local environment. LGTM
> 
> I asked Sophie to do a quick pull on her end from upstream/master so this should be resolved!

Ok sounds good!

PR #186, opened 2020-03-04
Remove erroneous article data from learn page

jestapinski:
This PR fixes the blank article on the learn page. What happened is we were getting data from stitch that looked like:

```
{ _id: 'some id',
  query_fields: {
    slug: '/'
    // no other fields
  }
}
```

While this data should be removed from stitch, we can also filter it out at build time. No learn page article should link to the home page.


jestapinski:
Thanks everyone!

PR #187, opened 2020-03-04
move title

greggb:
Could use a hand checking for any places that still use children instead of title where appropriate. Think I caught all the cases.

We should use title for text only titles and only pass children for things like the twitch feature.

This also bumps the featured home page articles to 3 lines which is different than the design, but prevent the cutoff on most of the articles.


greggb:
thanks!

PR #193, opened 2020-03-04
Filter Without Stitch, Dynamically Change Counts

jestapinski:
This PR adds build-time logic to compute filtering values for the learn page filters. Specifically, now:

- Choosing a filter dynamically updates the other filter to show options where at least one article exists.
- Choosing filters dynamically changes the number shown available in the other filter.
- **Choosing both filters removes count numbers from the filters** (keeping numbers here did not make sense).

Please give this a try locally and see if anything is wonky!

The logic is a bit odd but I think this best supports how we want to use the filters. I chose to represent the filters like so:

Screen Shot 2020-03-04 at 5 19 40 PM

Where each value loooks something like (here for `languages`):
Screen Shot 2020-03-04 at 5 21 30 PM

### DIFF
--- a/src/components/dev-hub/select.js
+++ b/src/components/dev-hub/select.js
@@ -78,6 +78,7 @@ const FormSelect = ({
     errors = [],
     narrow = false,
     onChange = null,
+    styleSelectedText = null,
     validationStatus = null,
     value = '',
     ...extraProps
@@ -89,6 +90,15 @@ const FormSelect = ({
         setShowOptions(!showOptions);
     }, [showOptions]);
     const selectOptions = typeof choices !== 'undefined' ? choices : children;
+    const updateSelectedText = useCallback(
+        text => {
+            const updatedText = styleSelectedText
+                ? styleSelectedText(text)
+                : text;
+            setSelectText(updatedText);
+        },
+        [styleSelectedText]
+    );
     /**
      * This useEffect should only be called once the component first renders with choices,
      * this should populate the select item with the default choice if there is one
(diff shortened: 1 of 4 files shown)


melissarudofsky:
it looks like after setting the filters, the numbers disappear 
Screen Shot 2020-03-05 at 9 59 49 AM

melissarudofsky:
not part of this PR but we have javascript, JavaScript, and Javascript

jestapinski:
> it looks like after setting the filters, the numbers disappear
> Screen Shot 2020-03-05 at 9 59 49 AM

Yeah let me double check this logic. If two filters are active I wanted to hide the numbers because they might not make sense with each other.

src/components/dev-hub/select.js
```diff
@@ -78,6 +78,7 @@ const FormSelect = ({
     errors = [],
     narrow = false,
     onChange = null,
+    styleSelectedText = null,
```
jestapinski: This prop helps add additional formatting to the selected text, so we can provide it different styling (such as removing the number) if need be. This addresses the hiding numbers issue in the dropdown.

src/components/dev-hub/select.js
```diff
@@ -45,7 +45,7 @@ const Options = styled('ul')`
         narrow
             ? `${OPTIONS_POSITION_OFFSET_NARROW}px`
             : `${OPTIONS_POSITION_OFFSET}px`};
-    width: 100%;
+    width: calc(100% + ${2 * BORDER_SIZE}px);
```
jestapinski: The dropdown did not seem to expand all the way across the component. I'll look into other select uses to make sure this aligns.

melissarudofsky:
testing it out locally and it seems like the numbers aren't matching up. could it be related to the blank article card?
Screen Shot 2020-03-05 at 1 58 50 PM
Screen Shot 2020-03-05 at 1 59 06 PM

jestapinski:
Yeah @melissarudofsky can re-start the dev server to make new pages?

jestapinski:
Thanks!

PR #198, opened 2020-03-05
Remove unused large images

jestapinski:
We have tons of large image files we are not actively using in the code, so this PR removes them to slim down the repo size.


jestapinski:
Thanks!

PR #200, opened 2020-03-06
Open twitch modal on homepage instead of link out

jestapinski:
This PR switches the "watch" button on the homepage to open a modal to play a video in-site instead of linking out.


jestapinski:
Thanks!

PR #203, opened 2020-03-06
Add Twitch URL to Home Page Notification

jestapinski:
This PR resolves the undefined URL issue for the live notification on the home page. We did not get a url from twitch on the stream, so here we just add the channel url in.


jestapinski:
Thanks!

PR #206, opened 2020-03-06
Align copy icon and text on copy button

jestapinski:
This PR fixes some alignment and padding issues with the copy button on codeblocks to address design concerns:

Screen Shot 2020-03-06 at 3 38 35 PM


jestapinski:
Thanks!

PR #207, opened 2020-03-06
Remove hover from secondary buttons

jestapinski:
This PR removes the slide-out hover state from secondary buttons, but keeps the existing border change on hover.

There is a primary button with the hover issue on the community page, but not sure how we would like to add a border to it. We may need a design for border, etc to remove the slide-out from primary buttons as well.

### DIFF
--- a/src/components/dev-hub/button.js
+++ b/src/components/dev-hub/button.js
@@ -37,6 +37,7 @@ const secondaryHoverStyles = css`
     &:active,
     &:hover,
     &:focus {
+        color: ${colorMap.devWhite};
         border: 2px solid ${colorMap.lightGreen};
     }
 `;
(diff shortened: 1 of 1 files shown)


jestapinski:
> Thanks for tackling this. Should we remove the hover effect on the primary button as well? We may be coming up with new hover effects for both.

Good point, I think we should as well. The issue is there is no current border on the primary button, so there would be no hover state without the slide. I can reach out to Chris to help with figuring out a proper hover state should we want to do this.

jbouzi12:
> > Thanks for tackling this. Should we remove the hover effect on the primary button as well? We may be coming up with new hover effects for both.
> 
> Good point, I think we should as well. The issue is there is no current border on the primary button, so there would be no hover state without the slide. I can reach out to Chris to help with figuring out a proper hover state should we want to do this.

That makes sense. Let's leave it in then until we can get a new animation, because some pages still display this animation for the primary button.

PR #220, opened 2020-03-09
Gb/updated images

greggb:
Two things in here. 

- Added all the new content cards (hoping they are small enough to leave them all in the repo and not choke Gatsby)
- Show a larger preview in the modal for twitch embeds and accept a preview thumbnail in the component.


greggb:
Actually looks like something funky is happening with the homepage feature image sizes. need to investigate

PR #221, opened 2020-03-10
Jb/events sorting

jbouzi12:
This PR updates all our events lists to ensure:
- they are sorted
- they only display upcoming events (I defined this as today or later)

### DIFF
--- a/src/components/dev-hub/event-list.js
+++ b/src/components/dev-hub/event-list.js
@@ -39,19 +39,16 @@ const CenterBlock = styled('div')`
 `;
 
 export const EventsListPreview = () => {
-    // TODO: Update implementation below
-    // to handle load/error states when events api CORS issue resolved
     const [events, error] = useEventData(EVENTS_API);
     const previews = events ? events.slice(0, 3) : [];
 
-    return previews.length ? (
+    return (
         <EventsPreview>
-            {previews.map(event => (
-                <Event key={event.url} event={event} />
-            ))}
+            {!error &&
+                previews.length &&
+                previews.map(event => <Event key={event.url} event={event} />)}
+            {error && <P>Check back later for upcoming events</P>}
         </EventsPreview>
-    ) : (
-        <P>Check back later for upcoming events</P>
     );
 };
 
--- a/src/components/dev-hub/event-list.js
+++ b/src/components/dev-hub/event-list.js
@@ -2,11 +2,12 @@ import React, { useEffect, useState } from 'react';
 import styled from '@emotion/styled';
 import Button from './button';
 import Event from './events';
+import { P } from './text';
 import useEventData from '../../hooks/use-event-data';
 import { size, screenSize } from './theme';
 
 export const EVENTS_API =
-    'https://www.mongodb.com/api/event/all/1?sort=-created_at&populate=tag_ids,node_ids';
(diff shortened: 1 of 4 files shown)


src/hooks/use-event-data.js
```diff
@@ -49,6 +49,21 @@ export const sampleEvents = [
     },
 ];
 
+export const removePastEvents = events => {
```
jestapinski: Super nit, but I noticed the `React` import on line 1 is no longer used. Should we remove it?

jbouzi12: Good catch 👍

jbouzi12:
> LGTM, I took a look in Firefox and Edge to make sure everything was filtered appropriately and looked fine.
> 
> Not sure if there is any concern with UTC comparison like in the author off by one date issue. It looks like the dates in the sample are timezone aware and the compared date (today) is UTC, so maybe there could be an off by one issue? That might be worth a look, but not really sure.
> 
> Looks great!

One thing is that this sorting wont work on IE because it doesn't support `Array.sort()`. But I assumed that we won't be supporting IE because our target audience is much less likely to be on it.

The date disparity is a good point though. I'm going to update the comparison to ensure all dates used are in `UTC`.

src/hooks/use-event-data.js
```diff
@@ -49,6 +49,21 @@ export const sampleEvents = [
...
+    return events.filter(
+        e => new Date(e.node_type_attributes.event_end) >= today
+    );
+};
+export const sortEvents = events =>
+    // create new list first so original does not get mutated
+    // this improves testability and reduces side-effects
+    [...events].sort((a, b) => {
+        const date1 = new Date(a.node_type_attributes.event_start).getTime();
+        const date2 = new Date(b.node_type_attributes.event_start).getTime();
```
greggb: Is the `getTime` necessary?

src/hooks/use-event-data.js
```diff
@@ -49,6 +50,23 @@ export const sampleEvents = [
...
+    return events.filter(
+        e =>
+            createDateObject(new Date(e.node_type_attributes.event_end)) >=
+            today
+    );
+};
+export const sortEvents = events =>
+    // create new list first so original does not get mutated
+    // this improves testability and reduces side-effects
+    [...events].sort((a, b) => {
```
greggb: I knew we had spoken briefly about this, but seems like we could get sorted events with ` https://www.mongodb.com/api/event/all/1?sort=-node_type_attributes.event_start`

src/hooks/use-event-data.js
```diff
@@ -49,6 +50,23 @@ export const sampleEvents = [
     },
 ];
 
+export const removePastEvents = events => {
+    const today = createDateObject(new Date());
+    return events.filter(
+        e =>
+            createDateObject(new Date(e.node_type_attributes.event_end)) >=
```
greggb: We should also filter on `status === 'published'`

src/components/dev-hub/event-list.js
```diff
@@ -49,7 +50,9 @@ export const EventsListPreview = () => {
                 <Event key={event.url} event={event} />
             ))}
         </EventsPreview>
-    ) : null;
+    ) : (
+        <P>Check back later for upcoming events</P>
```
greggb: Do we need this to match the error handling on the events page?

PR #222, opened 2020-03-10
Fix article dates off by one day

jestapinski:
This PR fixes article dates being shown one day earlier than they should be (for example, an article could be published on August 10th and show August 9th).

What was happening was we were using `toLocaleString` on a date object that was being interpreted as UTC, without specifying 'UTC' as the timezone option in `toLocaleString` (we were using the browser's timezone).

Before:
(The first date is the date the article was published, the second date was after running the date through the `createDateObject` helper, and the third was the return value of `toDateString` without the 'UTC' tz declared in options being passed).
Screen Shot 2020-03-10 at 10 02 59 AM

After:
Screen Shot 2020-03-10 at 10 02 41 AM

I pulled the hour and minutes out, but wanted to show for effect.

### DIFF
--- a/src/templates/article.js
+++ b/src/templates/article.js
@@ -35,6 +35,7 @@ const dateFormatOptions = {
     month: 'short',
     day: '2-digit',
     year: 'numeric',
+    timeZone: 'UTC',
 };
 
 /**


jestapinski:
Thanks!

PR #223, opened 2020-03-10
Gb/links and refs

greggb:
Fixes some issues with links and anchors

- All `RefRole` targets should use `to` in order for the `Link` component to format them correctly.
- doc roles (links to other articles) should check for and render any children before trying to use the slugMap for the link text


jbouzi12:
What would be the best way to test this out?

greggb:
> What would be the best way to test this out?

Ideally with the test page we talked about, but for now I tested it locally by copying the pageContext into my article component manually.

PR #226, opened 2020-03-10
Add more restrictive phantom data filtering

jestapinski:
This PR solves odd phantom data in the `learn` page.

What was happening is we were getting data from Stitch for articles that were oddly misshaped (see items 22 and 23):

Screen Shot 2020-03-10 at 2 15 27 PM

This PR adds filtering to look for additional fields before handing data off to the learn page. We were already doing this for `/` cards.


jestapinski:
Thanks!

PR #231, opened 2020-03-12
Update Twitter Share Text, Refactor Share URLs

jestapinski:
This PR addresses [DEVHUB-66](https://jira.mongodb.org/browse/DEVHUB-66) by adding some detailed Tweet text to the auto-populated Tweet when sharing an article. The new copy looks like the following:

Screen Shot 2020-03-12 at 12 04 23 PM

I also refactored the URLs between the article share footer and share tooltip so future copy changes can be easier. I also asked about updating copy for other Social Media choices as well, no responses yet there.

### DIFF
--- a/src/utils/get-article-share-links.js
+++ b/src/utils/get-article-share-links.js
@@ -1,0 +1,8 @@
+export const getArticleShareLinks = (title, url) => {
+    const facebookUrl = `https://www.facebook.com/sharer/sharer.php?u=${url}`;
+    const twitterUrl = `https://twitter.com/intent/tweet?url=${url}&text=${encodeURIComponent(
+        `Here is a post from @mongodb on "${title}"`
+    )}`;
+    const linkedInUrl = `https://www.linkedin.com/shareArticle?url=${url}`;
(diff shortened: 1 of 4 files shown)


greggb:
Getting some feedback from writers on the copy

src/utils/get-article-share-links.js
```diff
@@ -0,0 +1,8 @@
+export const getArticleShareLinks = (title, url) => {
+    const facebookUrl = `https://www.facebook.com/sharer/sharer.php?u=${url}`;
+    const twitterUrl = `https://twitter.com/intent/tweet?url=${url}&text=${encodeURIComponent(
+        `Here is a post from @mongodb on "${title}"`
```
greggb: Let's go with just `title`. They're adding a new 'title' field to the twitter meta as well, but I don't think we have access to that data here.

jestapinski: So the text would be just `${title}` and nothing more?

jestapinski:
Thanks!

PR #233, opened 2020-03-13
add async await to authentication for the project form

melissarudofsky:


src/components/dev-hub/project-sign-up-form.js
```diff
@@ -56,15 +56,15 @@ const Form = React.memo(({ setSuccess, success }) => {
...
         e.preventDefault();
         setCanSubmit(false);
         const obj = {
             name,
             email,
             projectDescription,
         };
-        authenticate();
+        await authenticate();
         const callback = hasSuccess => {
```
jestapinski: Hmm just tried on chrome incognito and am still getting the submission failed. I looked in the network tab and the stitch calls looked fine, not sure what is going on here?

jestapinski: Well I re-started the development server and that seemed to work! It takes a second though, should we have some kind of loading state?

melissarudofsky: hmm good point. maybe the authentication should happen on mount instead then ?

jestapinski: I think it might have been something on my end?

melissarudofsky: oo i thought I saw a comment about a loadign state

jestapinski: Yeah I was trying to debug a bit but then restarting my environment worked so I updated that comment!

melissarudofsky: I think this way may be better anyway, because once you pointed that out, I did notice the delay in seeing the loading state. What do you think?

PR #241, opened 2020-03-16
WIP RSS Feed for Articles

jestapinski:
_There are some interesting gatsby tidbits that came up with this so marking as WIP to sort those out, but. feel free to leave other feedback as well_

This PR adds an rss feed to production builds of the devhub, via a new plugin for gatsby, where each article looks as such:
Screen Shot 2020-03-16 at 5 01 43 PM

Do we want to add more fields to the feed? I figured title, description, and link would suffice.

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -1,3 +1,4 @@
+const dlv = require('dlv');
 const { generatePathPrefix } = require('./src/utils/generate-path-prefix');
 const { getMetadata } = require('./src/utils/get-metadata');
 
@@ -32,6 +33,83 @@ module.exports = {
             options: {
                 id: 'GTM-GDFN',
                 includeInDevelopment: false,
+            },
+        },
+        {
+            resolve: `gatsby-plugin-feed`,
+            options: {
+                query: `
+                {
+                    site {
+                    siteMetadata {
+                        title
+                        siteUrl
+                    }
+                    }
+                }
+                `,
+                feeds: [
+                    {
+                        serialize: ({ query: { site, allSitePage } }) => {
+                            return allSitePage.edges.map(edge => {
+                                const path = dlv(edge, 'node.path');
+                                const query_fields = dlv(
+                                    edge,
+                                    'node.context._xrefDocMapping.query_fields'
+                                );
+                                const description = dlv(
+                                    query_fields,
+                                    [
+                                        'meta_description',
+                                        0,
+                                        'children',
+                                        0,
+                                        'value',
+                                    ],
+                                    'No description available'
+                                );
+                                const title = dlv(
+                                    query_fields,
+                                    ['title', 0, 'value'],
+                                    'No title available'
+                                );
+                                return {
+                                    title: title,
+                                    description: description,
+                                    url: site.siteMetadata.siteUrl + path,
+                                };
+                            });
+                        },
+                        query: `
+                    {
+                        allSitePage(filter: {path: {regex: "/^/((article)|(how-to)|(quickstart))/"}}) {
+                            edges {
+                                node {
+                                id
+                                path
+                                context {
+                                    _xrefDocMapping {
+                                        query_fields {
+                                            meta_description {
+                                                children {
+                                                    value
+                                                }
+                                            }
+                                            title {
+                                                value
+                                            }
+                                        }
+                                    }
+                                }
+                                }
+                            }
+                        }
+                    }
+            `,
+                        output: '/rss.xml',
+                        title: 'MongoDB Developer Hub RSS Feed',
+                    },
+                ],
             },
         },
     ],
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -1,4 +1,4 @@
-const dlv = require('dlv');
+const { getNestedValue } = require('./src/utils/get-nested-value');
 const { generatePathPrefix } = require('./src/utils/generate-path-prefix');
 const { getMetadata } = require('./src/utils/get-metadata');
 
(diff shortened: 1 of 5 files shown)


gatsby-node.js
```diff
@@ -281,6 +283,7 @@ exports.createPages = async ({ actions }) => {
                     slug,
                     snootyStitchId: SNOOTY_STITCH_ID,
                     __refDocMapping: pageNodes,
+                    _xrefDocMapping: pageNodes,
```
jestapinski: @sophstad I noticed something else interesting about gatsby. When querying in production (via graphql) for a page's "__refDocMapping", graphql for whatever reason when it creates a schema sets it to "_xrefDocMapping". Meaning when we query graphql directly (which we do for this plugin in `gatsby-config`), we have to query `_xrefDocMapping`, since it's part of the generated schema, but that isn't what the field is really referring to. When I added this line to explicitly set it, I had success.

I also got a warning at build time that this assignment was a duplicate, but the previous value was null from graphql. Any thoughts on the best way to move forward with this?

gatsby-node.js
```diff
@@ -270,6 +270,8 @@ exports.createPages = async ({ actions }) => {
             if (pageNodes.query_fields) {
                 const relatedPages = getRelatedPagesWithImages(pageNodes);
                 pageNodes['query_fields'].related = relatedPages;
+                pageNodes['query_fields'].meta_description =
+                    pageNodes['query_fields']['meta-description'];
```
jestapinski: Similar note here than below for `_xrefDocMapping`

gatsby-node.js
jestapinski: When running on development, the same issue occurs. Here is the schema on development in graphiql.

Screen Shot 2020-03-16 at 5 21 02 PM

gatsby-config.js
```diff
@@ -27,6 +28,83 @@ module.exports = {
                 includeInDevelopment: false,
             },
         },
+        {
+            resolve: `gatsby-plugin-feed`,
```
greggb: I just merged a PR which uses the plugins from `gatsby-config` in `gatsby-config.prod.js` to prevent redundancy. This'll be a conflict now.

gatsby-config.js
```diff
@@ -27,6 +28,83 @@ module.exports = {
...
+                                    'node.context._xrefDocMapping.query_fields'
+                                );
+                                const description = dlv(
+                                    query_fields,
+                                    [
+                                        'meta_description',
+                                        0,
+                                        'children',
+                                        0,
+                                        'value',
```
greggb: Wes should use get-nested-value for this in case there are any intermediate nodes.

gatsby-config.js
```diff
@@ -27,6 +28,83 @@ module.exports = {
...
+                                    ],
+                                    'No description available'
+                                );
+                                const title = dlv(
+                                    query_fields,
+                                    ['title', 0, 'value'],
+                                    'No title available'
```
greggb: If we don't have a title or description we should leave these blank (''). If this shows up in my rss reader with 'No title available' it's kind of weird. It's much more common to just have it missing.

gatsby-node.js
greggb: Isn't this what we're using the stitch functions for, rather than querying gql? Can we use the same function we use to get all the `learn` articles to populate this list as well?

jestapinski: Yes, we do use the stitch functions to query this data. The `gatsby-plugin-feed` plugin, however, runs after the pages are built.

My understanding is we pull data from stitch to get page data, then for each page run the `createPage` method. After all of the pages are built, the inner graphql schema is updated to represent the page nodes as well, which is where these values are represented and generalized into an internal schema for graphql.

It looks like [here](https://www.gatsbyjs.org/docs/gatsby-lifecycle-apis/#bootstrap-sequence) we see the graphql schema being generated before pages are built, and then updated right after. The `gatsby-plugin-feed` config is run during the `onPostBuild` step at the very end.

Since at this point in the build cycle our schema should be complete, the plugin expects graphql to be the source of truth for pages and **requires** a graphql query to fetch page data.

Of course, we could try to do this without the plugin, but not sure that is worth the effort. I'm just confused why this field is being renamed in the schema but isn't a problem for pages.

greggb: > requires a graphql query to fetch page data

This isn't technically true, but is the preferred way.

I just don't see us using gql anywhere else on the site and use stitch as our source of truth. In the strictest technical sense we should be adding our stitch functions as a source plugin which would make the source of truth and gql match more closely.

jestapinski: okay! I can dig into this a bit more here

gatsby-config.js
```diff
@@ -34,6 +35,91 @@ module.exports = {
...
+                    siteMetadata {
+                        title
+                        siteUrl
+                    }
+                    }
+                }
+                `,
+                feeds: [
+                    {
+                        serialize: ({ query: { site, allSitePage } }) => {
```
greggb: Any thoughts on pulling some stuff out of here for testing/readability? I had to revert the plugins import change since we apparently only have access to the prod config in production. We might want to put these in a shared place.

jestapinski: that makes sense, I still want to revisit this a bit further as well

jestapinski:
I'm going to close this PR for now, because after some digging I think there is some prerequisite work we should do for this to support this plugin properly, mainly around how we can better use Gatsby's main concept of [nodes](https://www.gatsbyjs.org/docs/node-interface/) to support working with our data.

PR #242, opened 2020-03-17
Support Multiple Authors (and Authors as Array)

jestapinski:
_This PR must merge in sync with a docs platform autobuilder change_

This PR does two things:

- Allows support of the `author` value in `meta` to be an array of author objects, instead of a single author object.
- Adds styling to support articles with multiple authors at once:

One author:
Screen Shot 2020-03-17 at 4 49 26 PM

Two: 
Screen Shot 2020-03-17 at 4 49 16 PM

Three:
Screen Shot 2020-03-17 at 4 48 11 PM

### DIFF
--- a/src/components/dev-hub/byline-block.js
+++ b/src/components/dev-hub/byline-block.js
@@ -36,14 +36,33 @@ const StyledAuthorImage = styled(AuthorImage)`
     margin-right: ${size.small};
 `;
 
-const BylineBlock = ({ authorImage, authorName = '' }) => {
-    const authorLink = `/author/${getTagPageUriComponent(authorName)}`;
+const AuthorImages = ({ authors }) => (
+    <div>
+        {authors.map(({ name, image }) => (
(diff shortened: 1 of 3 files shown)


src/components/dev-hub/byline-block.js
```diff
@@ -17,9 +21,21 @@ const AuthorLink = styled(Link)`
 `;
 
 const AuthorText = styled(P)`
+    display: inline-block;
     @media ${screenSize.upToLarge} {
         font-size: ${fontSize.xsmall};
     }
+    &:before {
```
greggb: I'm concerned these pseudo-elements won't be picked up by search engines and screen readers. Can you add them directly to the text?

greggb:
I saw an array coming back locally, is the docs team ready with the parser?

jestapinski:
They have the array set up on local (test) data but have not released it yet to their production autobuilder. Sophie said she would tell me when they plan to do so.

PR #244, opened 2020-03-18
Support toml Featured Articles, Refactor gatsby-node

jestapinski:
This PR adds the ability to accept new arrays in the content repo's top-level [snooty.toml](https://github.com/mongodb/devhub-content/blob/master/snooty.toml) file as such:

```
[featured_articles]
"homePage" = ['path/to/article', ...]
"learnPage" = ['path/to/article', ...]
```

which allows the content team to dynamically change featured articles on the home and learn pages.

This PR. is just the changes to `gatsby-node` to pull this. data in and pass to the home/learn pages. A following PR ( #245 ) will make the change on the home page to show dynamic featured articles.

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -31,7 +31,14 @@ let PAGE_ID_PREFIX;
 const PAGES = [];
 const IMAGE_FILES = {};
 
-const FEATURED_LEARN_SLUGS = [
+const DEFAULT_FEATURED_HOME_SLUGS = [
+    '/how-to/nextjs-building-modern-applications',
+    '/how-to/python-starlette-stitch',
+    '/how-to/graphql-support-atlas-stitch',
+    '/quickstart/free-atlas-cluster',
+];
+
+const DEFAULT_FEATURED_LEARN_SLUGS = [
     '/quickstart/java-setup-crud-operations',
     '/how-to/golang-alexa-skills',
     '/how-to/polymorphic-pattern',
@@ -338,13 +345,16 @@ const getLearnPageArticles = async () => {
     return filteredDocuments;
 };
 
-const getFeaturedArticles = (allArticles, featuredArticles) => {
+const getFeaturedArticles = (allArticles, featuredArticleSlugs) => {
     const result = [];
-    featuredArticles.forEach((f, i) => {
-        const newArticle = allArticles.find(x => x.query_fields.slug === f);
+    featuredArticleSlugs.forEach((featuredSlug, i) => {
+        const newArticle = allArticles.find(
+            x => x.query_fields.slug === featuredSlug
+        );
         if (newArticle) {
             result.push(newArticle);
         } else {
+            // This article was not found. pick an existing article and add it instead.
             result.push(allArticles[i]);
         }
     });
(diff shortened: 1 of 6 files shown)


gatsby-node.js
```diff
@@ -401,21 +411,64 @@ const getLearnPageFilters = async () => {
...
+    defaultFeaturedArticles,
+    allArticles = null
+) => {
+    if (!allArticles) {
+        allArticles = await getLearnPageArticles();
+    }
+    // TODO: Verify field names with docsp
+    const featuredArticleSlugs =
+        getNestedValue(['featuredArticles', `${page}Page`], metadata) ||
+        defaultFeaturedArticles;
```
jestapinski: We can merge this change while the docs platform team is determining the values they want to support for this. If we don't get anything in metadata, we will use one of the `DEFAULT` slug arrays up top (as we are now for the learn page)

gatsby-node.js
```diff
@@ -401,21 +411,64 @@ const getLearnPageFilters = async () => {
...
+    return featuredArticles;
+};
+
+exports.onCreatePage = async ({
+    page,
+    actions: { createPage, deletePage },
+}) => {
+    switch (page.path) {
+        case '/learn/':
+            const allArticles = await getLearnPageArticles();
```
jestapinski: Is it worth looking into caching this?

greggb: I'm wondering if we should be doing this work is the same way we create the other dynamic pages. Instead of having a 'static' page in `pages` they would be templates.

With or without that change I wonder if there's an opportunity to ensure we're not making the same stitch calls more than once. We could store a map of returned calls and check there before reaching out to the server.

All this is done at build time, so I don't think it's a blocking issue (unless the build time becomes untenable), but this file is starting to get unwieldy and there's room for some cleanup and organization (and tests) now.

gatsby-node.js
```diff
@@ -401,21 +411,64 @@ const getLearnPageFilters = async () => {
...
+    page,
+    defaultFeaturedArticles,
+    allArticles = null
+) => {
+    if (!allArticles) {
+        allArticles = await getLearnPageArticles();
+    }
+    // TODO: Verify field names with docsp
+    const featuredArticleSlugs =
+        getNestedValue(['featuredArticles', `${page}Page`], metadata) ||
```
greggb: Should we move to `dlv` more consistently?

greggb: This might be too much of a nit, but `${page}Page` will shoe-horn us into that format, a map might be better here.

gatsby-node.js
```diff
@@ -401,21 +411,64 @@ const getLearnPageFilters = async () => {
...
-            ...page,
-            context: {
-                ...page.context,
-                allArticles,
-                featuredArticles,
-                filters,
-            },
-        });
+const getFeaturedArticlesForPage = async (
+    page,
```
greggb:
```suggestion
    pageName,
```

gatsby-node.js
jestapinski: > Should we move to `dlv` more consistently?

Yeah I was curious about this, since it is almost equivalent to getNestedValue. I wouldn't mind using `dlv` since we can specify a default without having to `|| <defaultValue>` which feels a bit. more hacky.

What do you think?

gatsby-node.js
jestapinski: Ah yes we can mimic our own cache here like that, sounds good!

I also agree on the need for cleanup. I was thinking we could move some of the lengthier API methods out (which would also allow them to be more testable). Now seems like a great time to do so, though, since more of the incoming features will use this file.

gatsby-node.js
greggb: Yep, I think that makes sense. Technically we could use optional changing which is built-in, but... then we'd have 3 ways to do this.

One thing to watch for when making this change is if we're getting a falsy value right now that then uses `||` to fallback. `dlv` I believe only fallsback when a value or key is undefined (like default params)

package.json
```diff
@@ -51,6 +51,7 @@
     "gatsby-plugin-google-tagmanager": "^2.1.25",
     "gatsby-plugin-react-helmet": "^3.1.14",
     "gatsby-plugin-sitemap": "^2.2.27",
+    "memoizerific": "^1.11.3",
```
jestapinski: We already had this as a dev-dependency, so I added it to our normal dependencies to use. It is only used at build-time. Was getting an `eslint(import/no-extraneous-dependencies)` warning, so figured this was the proper thing to do.

This package lets us memoize function calls, I was looking this to memoize stitch calls where possible.

src/utils/setup/on-create-page.js
```diff
@@ -0,0 +1,167 @@
...
+    // For each language, build an object with its total count, and count for each product
+    for (let i = 0; i < languageValues.length; i++) {
+        const l = languageValues[i];
+        filters.languages[l._id] = {
+            count: l.count,
+            products: {},
+        };
+        const productValuesForLanguage = await stitchClient.callFunction(
+            'getValuesByKey',
+            [metadata, 'products', { languages: l._id }]
```
jestapinski: I would also like to replace these stitch calls with some local filtering, but in a separate PR. There is some refactoring work we can do here to minimize the network I/O here but would like to do that separately.

src/utils/setup/on-create-page.js
```diff
@@ -0,0 +1,167 @@
...
+    const featuredArticles = getFeaturedArticles(
+        allArticles,
+        featuredArticleSlugs
+    );
+    return featuredArticles;
+};
+
+const onCreatePage = async (page, actions, inheritedStitchClient) => {
+    const { createPage, deletePage } = actions;
+    stitchClient = inheritedStitchClient;
```
jestapinski: Didn't want to keep passing the stitch client around :)

jestapinski:
This PR has grown a bit, since I moved out two large functions from `gatsby-node` into their own files. These are left almost unchanged, but there is some opportunity to refactor them soon.

I added some memoization for repeated stitch queries, but we can do some more going forward.

src/utils/setup/on-create-page.js
```diff
@@ -0,0 +1,167 @@
...
+        } else {
+            // This article was not found. pick an existing article and add it instead.
+            result.push(allArticles[i]);
+        }
+    });
+    return result;
+};
+
+const getLearnPageFilters = async () => {
+    const filters = {
```
greggb: Unless I'm missing something we could just assign two `const`s for languages and products and return the and object at the end of the function like `return {languages, products}`. This way we don't have to keep enlarging a nested object or access nested properties in these loops.

src/utils/setup/on-create-page.js
greggb: Yep, this'll make a huge improvement to build times

src/utils/setup/create-tag-page-type.js
```diff
@@ -0,0 +1,110 @@
...
+    // Once requests finish, map the item with name (and optional image) to the response's return value
+    const itemsWithPageData = possibleTagValues.map((r, i) => ({
+        item: r,
+        pages: pageData[i],
+    }));
+
+    const pageList = itemsWithPageData.map(page => {
+        const name = isAuthor ? page.item._id.name : page.item._id;
+        // Some bad data for authors doesn't follow this structure, so ignore it
+        if (!name) return null;
```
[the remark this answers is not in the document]
jestapinski: Good point

jestapinski:
Agreed on the optimizations, let's do them separately from this PR. Thanks!

PR #245, opened 2020-03-18
Render Dynamic Featured Articles on Home Page

jestapinski:
This PR pulls out the hardcoded articles on the home page and adds the ability to render featured articles dynamically, similar to with the learn page.

### DIFF
--- a/src/utils/get-featured-card-fields.js
+++ b/src/utils/get-featured-card-fields.js
@@ -1,0 +1,30 @@
+import dlv from 'dlv';
+
+export const getFeaturedCardFields = article => {
+    if (!article) {
+        return {
+            image: null,
+            slug: null,
+            title: null,
+            description: null,
+            tags: null,
+        };
+    }
+    const query_fields = article.query_fields;
+    const image = query_fields['atf-image'];
+    const slug = query_fields.slug;
+    const title = dlv(query_fields, ['title', 0, 'value']);
+    const description = dlv(query_fields, [
+        'meta-description',
+        0,
+        'children',
+        0,
+        'value',
+    ]);
+    const tags = {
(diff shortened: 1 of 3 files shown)


src/utils/get-featured-card-fields.js
```diff
@@ -0,0 +1,30 @@
...
+            title: null,
+            description: null,
+            tags: null,
+        };
+    }
+    const query_fields = article.query_fields;
+    const image = query_fields['atf-image'];
+    const slug = query_fields.slug;
+    const title = dlv(query_fields, ['title', 0, 'value']);
+    const description = dlv(query_fields, [
```
greggb: Can this use get-nested-value?

src/utils/get-featured-card-fields.js
```diff
@@ -0,0 +1,30 @@
...
+        'children',
+        0,
+        'value',
+    ]);
+    const tags = {
+        products: query_fields.products,
+        tags: query_fields.tags,
+        languages: query_fields.languages,
+    };
+    return { image, slug, title, description, tags };
```
greggb: What are your thoughts  on skipping creating `const`s and just assigning the values here in the return object?

src/utils/get-featured-card-fields.js
jestapinski: Is there a specific rule for using `getNestedValue` vs `dlv`? If it is a node should we default to `getNestedValue`?

greggb: Yea, that's what I'm thinking. There is a chance for there to be intermediate nodes we don't expect if there is a formatting error in the RST or a writer gets fancy.

jestapinski:
Thanks!

PR #246, opened 2020-03-18
Jb/update storybook theme

jbouzi12:
This PR is the first step to customizing and updating the storybook config. I initially planned to include the configuration updates but with shifting priorities I wanted to at least get the theme updates out there so the branch doesn't get stale.

### Future updates
- update prettier configuration to properly lint `mdx`
- update storybook configuration to fix emotion styled component parsing errors


.storybook/preview-head.html
```diff
@@ -0,0 +1,12 @@
+<style>
```
greggb: What area of the storybook do these styles apply to?

.storybook/preview-head.html
```diff
@@ -0,0 +1,12 @@
...
+    body {
+        font-family: 'Fira Mono', monospace;
+        color: '#fff';
```
greggb: I think the storybook html/body styles should match what we have in `layout.js`

This color isn't working because of the quotes.

.storybook/preview-head.html
```diff
@@ -0,0 +1,12 @@
...
+    }
+    .sbdocs-p code,
+    .sbdocs-li code {
+        color: #fff !important;
```
greggb: I don't think we use pure white anywhere on the site. The only exception I see is in `link.js` where I think I had a placeholder color that never got updated. I see a bunch of `#fff` in the SB theme, but that doesn't reflect the color we use.

.storybook/theme.js
```diff
@@ -0,0 +1,36 @@
...
+    appBorderRadius: 4,
+
+    // Typography
+    fontBase: `akzidenz, -apple-system, BlinkMacSystemFont, "Segoe UI",
+            Helvetica, Arial, sans-serif, 'Apple Color Emoji', "Segoe UI Emoji",
+            "Segoe UI Symbol"`,
+
+    // Text colors
+    textColor: '#fff',
+    textInverseColor: colorMap.devWhite,
```
greggb: How does this work? These two values are both white, what makes one 'inverse'? I think we should avoid #fff if possible since it's not part of the theme.

.storybook/preview-head.html
```diff
@@ -0,0 +1,12 @@
+<style>
+    body {
+        font-family: 'Fira Mono', monospace;
```
greggb: If this is part of the preview area shouldn't it be populated from theme file configs?

.storybook/theme.js
```diff
@@ -0,0 +1,36 @@
...
+    colorSecondary: colorMap.teal,
+
+    // UI
+    appBg: colorMap.pageBackground,
+    appContentBg: colorMap.devBlack,
+    appBorderColor: colorMap.greyLightOne,
+    appBorderRadius: 4,
+
+    // Typography
+    fontBase: `akzidenz, -apple-system, BlinkMacSystemFont, "Segoe UI",
```
greggb: I don't think akzidenz or fira font files are being loaded.

src/components/dev-hub/stories/theme.stories.mdx
```diff
@@ -42,8 +43,9 @@ import { colorMap, gradientMap, size, fontSize } from '../theme';
     <Story name="gradients">
         {() => {
             const Wrapper = styled('div')`
-                fontfamily: 'Fira Mono', monospace;
-                fontsize: ${fontSize.xsmall};
+                color: #fff;
```
greggb: Same comments as above, but I'm curious why we need to define these styles in multiple places. Shouldn't the theme be global for storybook?

jbouzi12:
@greggb Due to the issues with `mdx` not fully rendering emotion styled components properly (this item is a TODO for the storybook setup), certain styles aren't being rendered properly. For example text colors in some components and the font-family of some components aren't rendering properly. 

The `.storybook/theme.js` styles the majority of the storybook theme, but certain defaults (particularly the font styles) aren't getting to the styled components. For this reason I temporarily added font-styles to some stories themselves so they can at least be more visually correct in the mean time.

Also, I mainly used the styles in `preview-head.html` to style storybook's default inline code-blocs  and also attempt to apply the font-color and font-family globally through the `body` tag. But since these two styles still aren't hitting the styled components I'll take these body styles out.

greggb:
I found a couple issues around that - do either of these help?
https://github.com/storybookjs/storybook/issues/9602
https://github.com/storybookjs/storybook/issues/7325

jbouzi12:
Yea I found these before going back to the cert editor page work as they seemed promising. I have a separate branch where I started implementing some of these suggestions.

jbouzi12:
Thanks!

PR #250, opened 2020-03-19
Fix Twitch Video Size on Mobile Home Page

jestapinski:
On mobile the Twitch video on the home page is very tiny. Specifying the width to be 100% fixes this issue and allows users to see the video on mobile.

Before:
Screen Shot 2020-03-19 at 3 03 40 PM

After:
Screen Shot 2020-03-19 at 3 03 25 PM


jestapinski:
Thanks!

PR #257, opened 2020-03-20
Add pre-commit hooks for linting/testing

jestapinski:
Now that `npm test` works again, this PR enables linting and testing on commits (similar to with university) using husky/lint-staged.

After enabling this, there were some lint errors to handle, so these were taken care of as well.

### DIFF
--- a/preview-start.js
+++ b/preview-start.js
@@ -1,6 +1,6 @@
+/* eslint-disable import/no-unresolved */
 import React from 'react';
 import ReactDOM from 'react-dom';
-// eslint-disable-next-line import/no-unresolved
 import { getPageData } from 'previewSetup'; // Alias found in gatsby-node and webpack.config.js
 // Layouts
 import Layout from './src/components/dev-hub/layout';
--- a/src/components/dev-hub/controlled-tooltip.js
+++ b/src/components/dev-hub/controlled-tooltip.js
@@ -1,3 +1,4 @@
+/* eslint-disable no-unused-vars */
 import React from 'react';
 import styled from '@emotion/styled';
 import { css } from '@emotion/core';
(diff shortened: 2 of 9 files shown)


preview-start.js
```diff
@@ -1,6 +1,6 @@
+/* eslint-disable import/no-unresolved */
```
jestapinski: The `guides` and `guides-index` are not in this repo, so perhaps we should omit them here. I am not too familiar with the preview work, however.

greggb: I'm hazy on this as well. We'll probably need to use our own templates, but there is still work to be done on the docs platform side as well

src/components/dev-hub/controlled-tooltip.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-unused-vars */
```
greggb: What's the unused var in this file?

tests/unit/use-event-data.test.js
```diff
@@ -41,6 +41,7 @@ const timezonedEvents = [
 ];
 it('should remove events that have past', () => {
     const filtered = removePastEvents(events);
+    // eslint-disable-next-line no-unused-vars
```
greggb: Should we just not assign a variable below?

src/components/dev-hub/controlled-tooltip.js
jestapinski: There were a few props in the component not being used. I saw a lengthy comment about how this component was used and decided it was safer just to add the comment. We can remove them I have no issue with that.

tests/unit/use-event-data.test.js
jestapinski: Fair point

src/components/dev-hub/controlled-tooltip.js
greggb: I think they may have a different lint setup where you can't pull out spread items without the linter warning - I'm fine either way

jestapinski: I just went ahead and pulled them out

jestapinski:
Thanks!

PR #258, opened 2020-03-23
run lint and format on changed files

greggb:
Didn't see a need to leave the checking without formatting, so I simplified the npm scripts for those commands.

Also added the `--` before passing args to the npm script in the list-staged config. That will enable additional flags to be passed.


lint-staged.config.js
```diff
@@ -1,10 +1,10 @@
 module.exports = {
-    '*.{js}': [
-        'npm run lint:fix',
-        'npm run format:fix',
+    '*.{js, mdx, md}': [
+        'npm run lint',
+        'npm run format',
         'git add',
-        'npm test --bail --findRelatedTests --coverage=false ',
+        'npm test -- --bail --findRelatedTests --coverage=false ',
```
jestapinski: 🤦‍♂

PR #265, opened 2020-03-23
Refactor and Test onCreatePage

jestapinski:
This PR does a few things:

- Refactors logic to generate the filters for the learn page to not repeatedly call stitch. Instead we use the article objects we have to update language/product values.
- Cleans up var names from a prior PR
- Adds testing for both generating the filters and getting featured articles
- Adds a `utils` subdirectory in `/tests` with a simple config (since this is mainly for js helpers, not components).

### DIFF
--- a/src/utils/setup/on-create-page.js
+++ b/src/utils/setup/on-create-page.js
@@ -57,11 +57,10 @@ const getLearnPageFilters = async () => {
     const allArticles = await getAllArticles();
 
     allArticles.forEach(a => {
-        const langs = a.query_fields.languages;
-        const prods = a.query_fields.products;
-        // Just lang
-        if (langs) {
-            langs.forEach(l => {
+        const languagesForArticle = a.query_fields.languages;
+        const productsForArticle = a.query_fields.products;
+        if (languagesForArticle) {
+            languagesForArticle.forEach(l => {
                 if (languages[l]) {
                     languages[l].count++;
                 } else {
@@ -70,20 +69,17 @@ const getLearnPageFilters = async () => {
                         products: {},
                     };
                 }
-                if (prods) {
-                    prods.forEach(p => {
-                        if (languages[l].products[p]) {
-                            languages[l].products[p]++;
-                        } else {
-                            languages[l].products[p] = 1;
-                        }
+                if (productsForArticle) {
+                    productsForArticle.forEach(p => {
+                        languages[l].products[p]
+                            ? languages[l].products[p]++
+                            : (languages[l].products[p] = 1);
                     });
                 }
             });
         }
-        if (prods) {
-            // Just products
-            prods.forEach(p => {
+        if (productsForArticle) {
+            productsForArticle.forEach(p => {
                 if (products[p]) {
                     products[p].count++;
                 } else {
@@ -92,19 +88,16 @@ const getLearnPageFilters = async () => {
                         languages: {},
                     };
                 }
-                if (langs) {
-                    langs.forEach(l => {
-                        if (products[p].languages[l]) {
-                            products[p].languages[l]++;
-                        } else {
-                            products[p].languages[l] = 1;
-                        }
+                if (languagesForArticle) {
+                    languagesForArticle.forEach(l => {
+                        products[p].languages[l]
+                            ? products[p].languages[l]++
+                            : (products[p].languages[l] = 1);
                     });
                 }
             });
         }
     });
-
     return { languages, products };
 };
 
--- a/src/utils/setup/on-create-page.js
+++ b/src/utils/setup/on-create-page.js
@@ -150,4 +150,4 @@ const onCreatePage = async (
     }
 };
 
-module.exports = { getFeaturedArticles, getLearnPageFilters, onCreatePage };
+module.exports = { findArticlesFromSlugs, getLearnPageFilters, onCreatePage };
--- a/src/utils/setup/on-create-page.js
+++ b/src/utils/setup/on-create-page.js
@@ -150,4 +150,4 @@ const onCreatePage = async (
     }
 };
 
-module.exports = { getLearnPageFilters, onCreatePage };
+module.exports = { getFeaturedArticles, getLearnPageFilters, onCreatePage };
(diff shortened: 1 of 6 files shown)


src/utils/setup/on-create-page.js
```diff
@@ -56,54 +56,57 @@ const getFeaturedArticles = (allArticles, featuredArticleSlugs) => {
...
+                    productsForArticle.forEach(p => {
+                        languages[l].products[p]
+                            ? languages[l].products[p]++
+                            : (languages[l].products[p] = 1);
+                    });
+                }
+            });
+        }
+        // Go through products for this article and update filter info
+        if (productsForArticle) {
```
jestapinski: There is some logical redundancy between this conditional and the one on line 67, and so I tried to refactor into a single helper. This got quite complicated and much harder to understand, so I left these separate.

jbouzi12: Yea I see how these are very similar but the fields within the resulting data structures are where things can get hairy when trying to combine these. Is there a way we can unify the naming for the resulting objects in a way that makes sense and allow us to create a single helper?

jestapinski: That's a good point, perhaps something like:
```
languages: {
 count: 1,
 otherTagCounts: {
   stitch: 1,
 }
}
```

The biggest con here is this is clear now, but if we add a third filter it would be a bit harder to understand. I think there might be a way to ratify some of the difference here just not sure what the best way forward is, any thoughts?

jbouzi12: Yea that makes sense. Hmm I'll have to think about this a little more, especially considering the additional filters the scenario you raised. I'm ok with this moving forward for now though.

src/utils/setup/on-create-page.js
```diff
@@ -56,54 +56,57 @@ const getFeaturedArticles = (allArticles, featuredArticleSlugs) => {
...
-            'getValuesByKey',
-            [metadata, 'languages', { products: p._id }]
-        );
-        languageValuesForProduct.forEach(lp => {
-            filters.products[p._id].languages[lp._id] = lp.count;
-        });
-    }
-    return filters;
+    allArticles.forEach(a => {
+        const languagesForArticle = a.query_fields.languages;
```
greggb: I think it'd be better to expand these single letter vars
a => article
l => language or lang
p => product

src/utils/setup/on-create-page.js
```diff
@@ -56,54 +56,57 @@ const getFeaturedArticles = (allArticles, featuredArticleSlugs) => {
...
+        const productsForArticle = a.query_fields.products;
+        // Go through languages for this article and update filter info
```
greggb: Let's add a little more to this like what filter info we're including (count etc)

src/utils/setup/on-create-page.js
greggb: I'm happy with whatever direction y'all decide, but I'm okay with a bit of redundancy for clarity/extensibility here. As long as it's still straightforward to test.

tests/utils/get-featured-articles.test.js
```diff
@@ -0,0 +1,45 @@
+import { getFeaturedArticles } from '../../src/utils/setup/on-create-page';
+
+it('should correctly find featured articles given a set of requested articles', () => {
```
greggb: Should we think about capping the list to 4 and 3 for home and learn too?

src/utils/setup/on-create-page.js
```diff
@@ -56,54 +56,57 @@ const getFeaturedArticles = (allArticles, featuredArticleSlugs) => {
...
+                    languagesForArticle.forEach(l => {
+                        products[p].languages[l]
+                            ? products[p].languages[l]++
+                            : (products[p].languages[l] = 1);
+                    });
+                }
+            });
+        }
+    });
+    return { languages, products };
```
greggb: 👍

tests/utils/get-learn-page-filters.test.js
```diff
@@ -0,0 +1,85 @@
...
+        },
+        {
+            query_fields: {
+                languages: [],
+                products: ['server', 'stitch'],
+            },
+        },
+    ];
+
+    const result = getLearnPageFilters(allArticles);
```
greggb:
```suggestion
    const filterData = getLearnPageFilters(allArticles);
```

tests/utils/get-learn-page-filters.test.js
```diff
@@ -0,0 +1,85 @@
...
+    const languages = result.languages;
+    const products = result.products;
```
greggb: nit suggestion - check for result.language/products right in the expect rather than assigning. It'll draw a more clear line from the work to the test. I'd also move the `result` assignment right above the `expect`s

tests/utils/get-featured-articles.test.js
jestapinski: Yes we definitely should. The home page is flexible to support <= 4, but the learn page should really be 3 (otherwise it would break the design?).

jestapinski:
> Looks great - really nice refactor. Any stats on the build time increase/decrease after this? I had some suggestions on code style that might help.

Down by about ~2 seconds last time I checked. I'll check again after addressing the comments

src/utils/setup/on-create-page.js
jestapinski: I think at this point having the distinction between languages/products and products/languages supports keeping these a bit separate to understand. We can revisit if we find a great way to refactor, but I think this is the best balance of code clarity and testibility for now

src/utils/setup/on-create-page.js
```diff
@@ -50,54 +54,67 @@ const findArticlesFromSlugs = (allArticles, articleSlugs) => {
...
+                } else {
+                    // Update language reference directly in outer block object
+                    languages[language] = {
+                        count: 1,
+                        products: {},
+                    };
+                }
+                if (productsForArticle) {
+                    // If this article also has products, update those counts as well for this language
+                    const productsForLanguage = languages[language].products;
```
jestapinski: @greggb interesting note on these vars. Since we are updating destructively we have to leave off the last key to allow the operation to modify the referenced object and not just the scoped variable :)

jestapinski:
Thanks all!

PR #266, opened 2020-03-23
allow for centering images

greggb:


src/components/Image.js
```diff
@@ -5,8 +5,29 @@ import { css } from '@emotion/core';
 import { getNestedValue } from '../utils/get-nested-value';
 import { size } from './dev-hub/theme';
 
+const getAlignment = align => {
```
jestapinski: Can we also add this to the `Chart` component? There is similar logic there and they likely will want the same behavior between charts and images.

greggb:
verified the different alignment options in storybook, thanks

PR #271, opened 2020-03-24
Support Featured Articles in pageGroups

jestapinski:
This PR updates the logic for supporting featured articles to use the `pageGroups` field from Stitch's metadata. We still have a fallback case for if this has not been implemented yet by the authors or the parser.

Instead of diving into data from the `getMetadata` hook, we store the values from a single stitch call in gatsby-node and pass them through.

This will conflict with the refactor work, but I expect this PR to go in first, so I will handle conflicts on the refactor branch.


jestapinski:
Thanks!

PR #273, opened 2020-03-24
Pull saveAssetFiles, constructDbFilter Out of gatsby-node

jestapinski:
This PR pulls three more helper functions out of gatsby node, and adds testing for `constructDBFilter`. `saveAssetFiles` is a bit more difficult to test since it involves file I/O, so I left that out.


jestapinski:
Thanks!

PR #278, opened 2020-03-27
Account for possible no leading slash in passed data

jestapinski:
This PR allows the incoming featured articles from the parser to have an **optional** leading slash, where before it was mandatory.

### DIFF
--- a/src/utils/setup/on-create-page.js
+++ b/src/utils/setup/on-create-page.js
@@ -9,16 +9,16 @@ const MAX_LEARN_PAGE_FEATURED_ARTICLES = 3;
 const MAX_HOME_PAGE_FEATURED_ARTICLES = 4;
 
 const DEFAULT_FEATURED_HOME_SLUGS = [
-    '/how-to/nextjs-building-modern-applications',
-    '/how-to/python-starlette-stitch',
-    '/how-to/graphql-support-atlas-stitch',
-    '/quickstart/free-atlas-cluster',
+    'how-to/nextjs-building-modern-applications',
+    'how-to/python-starlette-stitch',
+    'how-to/graphql-support-atlas-stitch',
+    'quickstart/free-atlas-cluster',
 ];
 
 const DEFAULT_FEATURED_LEARN_SLUGS = [
-    '/quickstart/java-setup-crud-operations',
-    '/how-to/golang-alexa-skills',
-    '/how-to/polymorphic-pattern',
+    'quickstart/java-setup-crud-operations',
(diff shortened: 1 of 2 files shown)


src/utils/setup/on-create-page.js
```diff
@@ -42,7 +42,9 @@ const findArticlesFromSlugs = (allArticles, articleSlugs, maxSize) => {
     // If maxSize is undefined, this will return a shallow copy of articleSlugs
     articleSlugs.slice(0, maxSize).forEach((featuredSlug, i) => {
         const newArticle = allArticles.find(
-            x => x.query_fields.slug === featuredSlug
+            x =>
```
jestapinski: The tests caught an interesting case where the featured article slug was a substring of another article slug so i was not getting the correct response. Checking strict equality here is better than `indexOf` for that reason.

src/utils/setup/on-create-page.js
```diff
@@ -9,16 +9,16 @@ const MAX_LEARN_PAGE_FEATURED_ARTICLES = 3;
 const MAX_HOME_PAGE_FEATURED_ARTICLES = 4;
 
 const DEFAULT_FEATURED_HOME_SLUGS = [
-    '/how-to/nextjs-building-modern-applications',
-    '/how-to/python-starlette-stitch',
-    '/how-to/graphql-support-atlas-stitch',
-    '/quickstart/free-atlas-cluster',
+    'how-to/nextjs-building-modern-applications',
```
jestapinski: I also updated the default arrays to match our parser response.

tests/utils/find-articles-from-slugs.test.js
```diff
@@ -2,9 +2,9 @@ import { findArticlesFromSlugs } from '../../src/utils/setup/on-create-page';
 
 it('should correctly find featured articles given a set of requested articles', () => {
     let requestedFeaturedSlugs = [
-        '/quickstart/java-setup-crud-operations',
-        '/how-to/golang-alexa-skills',
-        '/how-to/polymorphic-pattern',
+        'quickstart/java-setup-crud-operations',
```
greggb: Could we test for a leading slash as well since the other series include them

jestapinski: Yep! We still are actually on line 36 here

greggb: Ah thanks, missed that!

jestapinski:
Thanks!

PR #280, opened 2020-03-27
Add npm test GH action

jestapinski:
This PR adds a simple action to run `npm test` on a PR to staging or master

### DIFF
--- a/.github/workflows/devhub-npm-test.yaml
+++ b/.github/workflows/devhub-npm-test.yaml
@@ -13,9 +13,9 @@ jobs:
 
         steps:
             - uses: actions/checkout@v2
-            - name: Use Node.js 13.x
+            - name: Use Node.js 12.x
               uses: actions/setup-node@v1
               with:
-                  node-version: '13.x'
+                  node-version: '12.x'
             - run: npm ci
             - run: npm test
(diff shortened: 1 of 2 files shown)


greggb:
Do the actions cache at all?

jestapinski:
> Do the actions cache at all?

[We can cache dependencies](https://help.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows) so those don't run each time. Should be pretty easy to add in here. Anything else we would want to cache?

greggb:
> > Do the actions cache at all?
> 
> [We can cache dependencies](https://help.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows) so those don't run each time. Should be pretty easy to add in here. Anything else we would want to cache?

Nothing that comes to mind

jestapinski:
Looks like we are hitting the cache and cut build times by ~10s, so seems to be working as expected for caching downloaded deps.

Screen Shot 2020-03-27 at 3 10 20 PM

greggb:
Where do you think the 37s are being spent in the `ci` task even with the cache?

jestapinski:
> Where do you think the 37s are being spent in the `ci` task even with the cache?

Yeah this seems a bit off, it looks like it is trying to install deps still but maybe not download. From the docs:

```
Jobs on GitHub-hosted runners start in a clean virtual environment and must download dependencies each time, causing increased network utilization, longer runtime, and increased cost. To help speed up the time it takes to recreate these files, GitHub can cache dependencies you frequently use in workflows.
```

So even with deps cached, it looks like it is still installing each time.

greggb:
Ha, that was the one, but looks like it does the same thing!

jestapinski:
🤷‍♂ It seems a bit quicker than before? Down to 52s from 59s, so it is worth keeping for now.

jestapinski:
Thanks!

PR #283, opened 2020-03-27
Account for og tag options in meta

jestapinski:
This PR adds in meta tags for `og` options based on information in an article from its `meta`. If these values are not supplied, we default to what we have been using (and added a default `atf-image` for `og:image`)

Screen Shot 2020-03-27 at 3 57 35 PM
Screen Shot 2020-03-27 at 3 57 42 PM


src/templates/article.js
```diff
@@ -177,10 +178,16 @@ const Article = props => {
...
             <Helmet>
                 <title>{articleTitle}</title>
-                <meta property="og:title" content={articleTitle} />
-                <meta property="og:url" content={articleUrl} />
-                <meta property="og:type" content="article" />
-                <link rel="canonical" href={articleUrl} />
+                <meta
+                    property="og:image"
+                    content={
+                        og.image || `${siteUrl}${withPrefix(meta['atf-image'])}`
```
greggb: I don't think we should fallback to the atf - they want to be specific with the image used for sharing. We *do* currently have a fallback in the `TwitterMeta` component to match the `og:` image with the `twitter:` image which should be sufficient if the `og:image` isn't specified. We'll need to ensure that the twitter component doesn't over-ride the `og:image` though

src/templates/article.js
```diff
@@ -177,10 +178,16 @@ const Article = props => {
...
+                    }
+                />
+                <meta property="og:title" content={og.title || articleTitle} />
```
greggb: Are the attributes definitely coming back in `meta`? The twitter directive gets passed through the factory.

src/templates/article.js
```diff
@@ -145,6 +145,7 @@ const Article = props => {
     const childNodes = dlv(__refDocMapping, 'ast.children', []);
     const contentNodes = getContent(childNodes);
     const meta = dlv(__refDocMapping, 'query_fields');
+    const og = meta.og || {};
```
greggb: We may need to create an SEO component (there is a gatsby doc on that) so we can consolidate all the `og` `twitter` `canonical` stuff

src/templates/article.js
jestapinski: They are, Sophie asked me where we would like to put them and I thought meta was best based on how we are using them now.

Screen Shot 2020-03-27 at 4 16 12 PM

src/templates/article.js
jestapinski: I'm assuming this can be done separately?

greggb: If all the varying requirements can be met in this PR without bringing them altogether, sure :)

greggb: Actually, looking at this, I think it would be better to move the twitter and og and canonical stuff into a component now so nothing is over-ridden. You'll also have to remove the twitter directive from the things we pass to the component factory.

jestapinski: Yeah I think now is the time to bring them all together.

jestapinski:
This PR has changed a bit to incorporate Twitter SEO data, canonical link use, and og SEO tags in one place (a new `SEO` component).

As a result, we pull `Twitter` out of `contentNodes` in the article to use here instead of in the ComponentFactory. Related Twitter components/ComponentFactory support have been removed since they should no longer be needed.

jestapinski:
Thanks! Will check through this one a bit more in staging

PR #284, opened 2020-03-30
Add loading state to events hook

jestapinski:
The community index page jumps a bit because it loads data from the events API. I noticed the community index page was missing a loading state, so I added it to the `useEventData` hook and copied the simple loading state from the events page into the `EventsListPreview` component

### DIFF
--- a/src/components/dev-hub/event-list.js
+++ b/src/components/dev-hub/event-list.js
@@ -13,6 +13,7 @@ const EventsPreview = styled('div')`
     display: flex;
     flex-direction: row;
     justify-content: space-between;
+    min-height: ${size.xxlarge};
     width: 100%;
     @media ${screenSize.upToMedium} {
         flex-direction: column;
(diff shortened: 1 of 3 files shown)


jestapinski:
> These changes look great, but when I went to the community and events pages I didn't see the loading state appear and I still noticed a jump when loading. Perhaps having a set height for the events preview would help as well

ah yeah perhaps it was super quick. I'll look into that change, thanks for the look!

src/components/dev-hub/event-list.js
```diff
@@ -13,6 +13,7 @@ const EventsPreview = styled('div')`
     display: flex;
     flex-direction: row;
     justify-content: space-between;
+    min-height: ${size.xxlarge};
```
jestapinski: This only solves it for the events preview on desktop. Mobile we could also do a min-height but since they are stacked it seemed a bit more natural?

jbouzi12: Yea that makes sense and is a good decision 👍

jestapinski:
Thanks!

PR #285, opened 2020-03-30
Js/design improvements

jestapinski:
This PR adds three design improvements for the devhub.

1. Fixes caption look, margin to match the prototype. The prototype used size 14 font and had the caption tucked against the image (I noticed we just were not using the prop for this on the `Image`.
2. Adds a collapse prop to remove the bottom border of the hero banner if needed, and uses this on the tag page.
3. Adds two hero banners for the tag pages, one for authors and one for all other tag pages. Kalie said the non-author one is just based off an existing background so I used that.

Author tag page:
Screen Shot 2020-03-30 at 4 31 22 PM

Non-author tag page:
Screen Shot 2020-03-30 at 4 31 33 PM


src/components/CaptionLegend.js
```diff
@@ -1,9 +1,14 @@
 import React from 'react';
+import styled from '@emotion/styled';
 import PropTypes from 'prop-types';
 import ComponentFactory from './ComponentFactory';
-import { P2 } from './dev-hub/text';
+import { P3 } from './dev-hub/text';
+import { colorMap } from './dev-hub/theme';
 
-const Caption = P2.withComponent('figcaption');
+const StyledCaption = styled(P3)`
```
melissarudofsky: are you using StyledCaption anywhere?

jestapinski: yep! Using it on line 11 to also use the `figcaption` tag

melissarudofsky: Are you trying to combine the styles of P3 with the color and then actually render a figcaption? I meant why can't like 8 just be `const Caption = styled('figcaption')`?

jestapinski: Ah yes that is what I am looking to do, would it make more sense just to bring the P3 styles into a `styled('figcaption')` here?

melissarudofsky: I guess this way is fine! just seemed weird to declare a style thats only used to declare another style but i think it does the trick!

jestapinski:
Thanks!

PR #286, opened 2020-03-31
Remove unused test json files

jestapinski:
We brought these test data files over from the Snooty repo, but are not using them in testing. This PR removes these unused JSON files.


jestapinski:
Thanks!

PR #287, opened 2020-03-31
Slim test json files and snapshots

jestapinski:
For test data files we are still using, there were some places where we were bringing in much more data than we really need. This PR slims these files, which also reduces the sizes of snapshots and cuts test time down by about half a second.

The bulk of this involved slimming data for the `LiteralInclude` test, which cut about 120,000 lines

There also was a PropType warning on the `RoleFile` test, I updated the proptype here to match the similar `RoleGuiLabel` to get rid of this.


jestapinski:
Thanks!

PR #288, opened 2020-03-31
Full-width Nav and Footer

jestapinski:
This PR updates the page layout to support a full-width nav and footer. The main logical change was moving `max-width` related info out of the global wrapper and into the nav, main, and footer, and constraining child elements to it.

The nav/footer changes are pretty similar in this way.

Other pages should look normal, I checked on Chrome/Firefox but some extra poking around the different pages just to confirm this would be great!

I also bumped the hover state color for the nav by request.

Dan is currently taking a look, but any changes forthcoming would likely be minor.

Screen Shot 2020-03-31 at 1 46 16 PM

### DIFF
--- a/src/components/dev-hub/global-footer.js
+++ b/src/components/dev-hub/global-footer.js
@@ -12,6 +12,9 @@ import Youtube from './icons/youtube';
 import Twitch from './icons/twitch';
 import { P } from './text';
 
+// Logo size 150px + 64px padding right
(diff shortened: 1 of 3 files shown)


jestapinski:
Thanks! After speaking with Dan we actually made the hover state a bit lighter so it doesn't confuse with the background:

Screen Shot 2020-04-01 at 10 11 37 AM

PR #290, opened 2020-04-01
Add versions.sh Script

jestapinski:
This PR adds a versions.sh script like we have for mitx to check staging against master. We can use this for deploys to the master branch in the future.

If you want to try locally, you will need to run `chmod 777 ./versions.sh` first to give execution permissions. There also isn't anything on staging not on master right now.

### DIFF
--- a/versions.sh
+++ b/versions.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+set -e
+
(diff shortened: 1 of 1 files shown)


jestapinski:
Thanks!

PR #291, opened 2020-04-02
Move env validation into onPreBootstrap part of the lifecycle

jestapinski:
I'm starting to refactor pieces of `gatsby-node` so we better match pieces of the Gatsby Node API based on what each Node API method is doing.

This piece moves some environment validation logic from the `sourceNodes` step of the lifecycle to the `onPreBootstrap` part of the lifecycle, because it is a better fit here. `sourceNodes` should simply be about pulling in data from any APIs, not validating env variables. `onPreBootstrap` runs earlier in the process and is for checking options and requirements.

As for the env validation itself, I decluttered an unnecessary return from it and instead just throw an error right in the function to abstract it from `gatsby-node`. I also added a quick test.

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -66,14 +66,9 @@ const getRelatedPagesWithImages = pageNodes => {
     return relatedPageInfo;
 };
 
+exports.onPreBootstrap = validateEnvVariables;
+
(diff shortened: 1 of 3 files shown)


jestapinski:
Thanks!

PR #292, opened 2020-04-02
Move Stitch Initialization to utils

jestapinski:
This PR starts some work on abstracting our data source (stitch) out of the gatsby node APIs. Here we just move the stitch setup into its own helper function and clean up the promise logic in it.


jestapinski:
Thanks!

PR #294, opened 2020-04-03
DOP-930: Add support for canonical-href option

sophstad:
[[JIRA](https://jira.mongodb.org/browse/DOP-930)] Requirements:
- If `:canonical-href: https://example.com` is specified for a page (on line 2, below `:template: devhub-article`), use this value as the page's canonical URL
- Otherwise, use the page's actual URL

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12640,25 +12640,25 @@
       "dependencies": {
         "abbrev": {
           "version": "1.1.1",
-          "resolved": "",
+          "resolved": false,
           "integrity": "sha512-nne9/IiQ/hzIhY6pdDnbBtz7DjPTKrY00P/zvPSm5pOFkl6xuGrGnXn/VtTNNfNtAfZ9/1RtehkszU9qcTii0Q==",
           "optional": true
         },
         "ansi-regex": {
           "version": "2.1.1",
-          "resolved": "",
+          "resolved": false,
           "integrity": "sha1-w7M6te42DYbg5ijwRorn7yfWVN8=",
           "optional": true
         },
         "aproba": {
           "version": "1.2.0",
-          "resolved": "",
+          "resolved": false,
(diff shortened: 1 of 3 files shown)


src/templates/article.js
```diff
@@ -179,11 +179,17 @@ const Article = props => {
         meta.updatedDate,
         dateFormatOptions
     );
+    const canonicalUrl = dlv(
+        __refDocMapping,
+        'ast.options.canonical-href',
+        null
```
jestapinski, line 185:
```suggestion
        og.url || articleUrl
```
We can actually set this to the fallback case for `dlv` and then remove the conditional in `SEO` since we can handle it here!

src/components/dev-hub/SEO.js
```diff
@@ -13,7 +13,15 @@ const getImageSrc = (src, siteUrl) => {
     return isExternalUrl.test(src) ? src : siteUrl + src;
 };
 
-const SEO = ({ articleTitle, ogTitle, image, url, type, twitterNode }) => {
+const SEO = ({
+    articleTitle,
+    canonicalUrl,
+    ogTitle,
+    image,
+    url,
```
jestapinski, line 21: Super nit, but perhaps now is a good time to rename `url` to `ogUrl` to be extra specific, since it would only be used for the `og:url` tag

package-lock.json
```diff
@@ -12640,25 +12640,25 @@
       "dependencies": {
         "abbrev": {
           "version": "1.1.1",
-          "resolved": "",
+          "resolved": false,
```
jestapinski, line 12643: I've also been noticing npm sneaking these changes into the lockfile on `npm install`. It isn't overriding any urls, just `""` so it should be safe to proceed with. I tried finding out more about this but nothing worth noting. We can check it in the staging branch just to be sure, but I was still able to freshly `npm install` locally without any issues.

PR #295, opened 2020-04-06
Pull twitter meta from childNodes directly

jestapinski:
This PR cleans up some twitter SEO logic, by just removing the `twitter` directive from the Article `contentNodesMap` and just checking the `childNodes` straight from the passed in data.

This allows us to remove the splice operation which was removing `twitter` from the contentNodes.


jestapinski:
Thanks!

PR #296, opened 2020-04-06
Fix MASTER to PROD for versions script

jestapinski:
The versions script was using `MASTER` instead of `PROD` for the branch alias, so it was not picking up the right branch to check


jestapinski:
Thanks!

PR #297, opened 2020-04-06
Extract Document Postprocessing

jestapinski:
This PR moves operations relating to postprocessing incoming documents from our data store (stitch) to a helper method with tests.

Doing so helps us more align with Gatsby API methodology by moving `onCreateNode` related processes to one place.

I also removed the unused `IMAGE_FILES` object.

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -2,6 +2,9 @@ const path = require('path');
 const dlv = require('dlv');
 const { constructDbFilter } = require('./src/utils/setup/construct-db-filter');
 const { initStitch } = require('./src/utils/setup/init-stich');
+const {
+    postprocessDocument,
+} = require('./src/utils/setup/postprocess-document');
 const { saveAssetFiles } = require('./src/utils/setup/save-asset-files');
 const {
     validateEnvVariables,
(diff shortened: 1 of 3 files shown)


gatsby-node.js
```diff
@@ -75,26 +77,18 @@ exports.sourceNodes = async () => {
...
-        if (key.includes('images/')) {
-            IMAGE_FILES[key] = val;
-        } else if (filename.endsWith('.txt')) {
-            PAGES.push(key);
-        }
+    documents.forEach(doc => {
+        // Mimics onCreateNode
+        const { page_id, ...nodeContent } = doc;
+        const slug = page_id.replace(`${PAGE_ID_PREFIX}/`, '');
+        postprocessDocument(
```
jestapinski, line 85: I chose to keep the in-memory array updates here to help this function mimic a potential [`onCreateNode` callback](https://www.gatsbyjs.org/docs/node-apis/#onCreateNode), should we move to using `createNode` to populate a data layer. This would be an easy switch.

greggb: What do you think about passing the `doc` to `postprocessDocument` and getting the `slug` and `nodeContent` in that function? Is that something that would clash with your goal of mimicing `onCreateNode`?

greggb: Or is this actually the cleaner approach? Hard to say with the different params passing around.

jestapinski: Yeah we can pull in the logic to get the slug. I was concerned about continuing to pass the `PAGE_ID_PREFIX`, but it would be nice to bring all the logic into one place.

jestapinski:
Thanks!

PR #298, opened 2020-04-06
Nav hover/active states, smaller font and height

jestapinski:
Dan has asked for the following changes to the nav:
- Remove hover/active states from the logo for home (existing convention should be enough for users to know it is clickable)
- Shrinks the size of the nav to 58px and 2px for the bar underneath
- Switches the hover/active backgrounds to use 40% opacity

Staging:
Screen Shot 2020-04-06 at 3 02 43 PM

This PR:
Screen Shot 2020-04-06 at 3 02 20 PM

Closer look at hover/active:
Screen Shot 2020-04-06 at 3 04 05 PM

### DIFF
--- a/src/components/dev-hub/global-nav.js
+++ b/src/components/dev-hub/global-nav.js
@@ -41,8 +41,8 @@ const NavLink = styled(Link)`
     text-decoration: none;
     &:hover,
     &[aria-current='page'] {
-        /* 66 is 40% opacity in hex */
-        background-color: ${`${colorMap.greyDarkTwo}66`};
+        /* greyDarkTwo at 40% opacity on greyDarkThree */
+        background-color: #2c3d47;
     }
     @media ${screenSize.upToMedium} {
         font-size: ${fontSize.tiny};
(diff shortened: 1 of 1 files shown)


src/components/dev-hub/global-nav.js
```diff
@@ -30,14 +33,16 @@ const NavContent = styled('div')`
...
-    line-height: ${lineHeight.xxlarge};
-    padding: ${size.small} ${size.xlarge};
+    line-height: ${lineHeight.small};
+    padding: ${LINK_VERTICAL_PADDING} ${size.xlarge};
     text-decoration: none;
     &:hover,
     &[aria-current='page'] {
-        background-color: ${colorMap.greyDarkTwo};
+        /* 66 is 40% opacity in hex */
+        background-color: ${`${colorMap.greyDarkTwo}66`};
```
greggb, line 45: we're not using hexa anywhere else, might be better to just get the resultant hex value.

jestapinski:
Thanks!

PR #299, opened 2020-04-06
Add Cypress Infrastructure For Local e2e

jestapinski:
This PR adds the basic cypress infrastructure to the devhub to allow us to begin writing tests:
- Added cypress and cypress-eslint-plugin to `package.json`
- Was running into eslint issues with `cy` not being defined, so added cypress to `.eslintrc`
- Added a basic rendering check for the home page (will replace later)

This PR does not add cypress to GitHub actions, I'm working with Sue and Sophie to figure out where we should pull data from for that.

To test locally, have `gatsby develop` running and then try `npm run test:e2e:start`. Cypress 4.3 also allows us to use Firefox or Edge too


jestapinski:
Thanks!

PR #301, opened 2020-04-08
Update useEventData hook to include live.mongodb.com events

jestapinski:
This PR adds data from the live.mongodb.com events API into the `useEventData` hook to populate the community and events pages. We also sort all events by their date, regardless of source. You can check it out now by going to `/community` or `/community/events`.

This approach is nice, because it keeps the page/event components blind to what the source of events data is, while we query multiple sources behind the scenes. We also are going to run into similar patterns in the future, so I would like to make sure this is the best way to handle these cases of pulling from multiple data sources for a single component.

**I was curious about a few things:**
- Does this approach make sense for handling multiple API calls in a single hook?
- Since we are not using TS yet, is there a better way to document needing to map one event object to fit the schema of another (we want all events to have a similar schema to use by the component, regardless of original source)


src/hooks/use-event-data.js
```diff
@@ -1,35 +1,26 @@
...
-                });
-                if (data) {
-                    const parsedData = await data.json();
-                    const upcomingEvents = removePastEvents(parsedData);
-                    setIsLoading(false);
-                    setError(null);
-                    setEvents(upcomingEvents.reverse());
-                }
+                const eventData = await fetchEventData(url);
+                const liveData = await fetchLiveEventData();
```
jbouzi12, line 15: I think this is a good pattern to use in this case because this hooks primary goal is to get events data. So it makes sense for us to get both regular and live events here.

src/utils/fetch-live-event-data.js
```diff
@@ -0,0 +1,28 @@
+const API_KEY = '00411c8fd1b9c3224260259d245dd29890130b94';
```
jbouzi12, line 1: Do we want the api-key hard coded here? Do we have a place where this can be stored?

src/utils/fetch-live-event-data.js
```diff
@@ -0,0 +1,28 @@
...
+    const data = await fetch(LIVE_EVENTS_URL, {
+        headers: {
+            Authorization: `Token ${API_KEY}`,
+            'Content-Type': 'application/json',
+        },
+    });
+    if (data) {
+        const parsedData = await data.json();
+        return parsedData.results.map(e => ({
+            // mongodb.com events api returns relevant fields in `node_type_attributes`
```
jbouzi12, line 15: Perhaps adding a little more context here to document the goal of keeping all events data consistent. Do we have any plans to add typescript at all?

src/utils/fetch-live-event-data.js
jestapinski: I asked Gregg about this and this key is public, so there is no worry about having it in the source code

src/utils/fetch-live-event-data.js
jestapinski: No plans yet for TS, but yeah maybe a comment or such would definitely help

src/hooks/use-event-data.js
```diff
@@ -1,35 +1,26 @@
...
-                    },
-                });
-                if (data) {
-                    const parsedData = await data.json();
-                    const upcomingEvents = removePastEvents(parsedData);
-                    setIsLoading(false);
-                    setError(null);
-                    setEvents(upcomingEvents.reverse());
-                }
+                const eventData = await fetchEventData(url);
```
greggb, line 14: Should we try to parallelize these requests with `Promise.all`?

src/utils/fetch-event-data.js
```diff
@@ -0,0 +1,33 @@
...
+            headers: {
+                'Content-Type': 'application/json',
+            },
+        });
+        if (data) {
+            const parsedData = await data.json();
+            const upcomingEvents = removePastEvents(parsedData);
+            return upcomingEvents;
+        }
+    } catch (e) {
```
jestapinski, line 27: One other piece I added in here was to make these fetch calls more robust, so still returning some array even if it errors out. This allows us to still display some events if the other source goes down or is inaccessible.

jestapinski:
> If we come up with a caching system for the video page we might want to extend that here. Some module that sits between the requests and the consuming component that checks if we've already fetched in the last minute or so.

Yeah I started thinking a bit about a general solution for this, perhaps some kind of `fetchWithCache` we can use in place. It seems the easiest solutions involve caching on the user's browser, would this route make sense?

jestapinski:
Thanks all! I'll look into the live response http caching separately

PR #302, opened 2020-04-08
Cache static assets to accelerate repeated builds

jestapinski:
Now that we have broken up pieces of gatsby-node, we can begin to optimize pieces of the build process by leveraging Gatsby's caching.

This PR optimizes writes to disk on repeated builds for static assets (images, etc.) and cuts build time by about 33% (about 4 seconds) on a repeated build (a build where the cache has already been populated).

What happens on a first build is the `createNode` method writes a new node to the cache which stores the asset ID in cached files. On a repeated build, we just use the cache instead of making a new node, therefore we do not call the `onCreateNode` handler and avoid writing out to disk.

I cleared my local `static` folder and re-ran the build. I didn't notice any missing assets.

### DIFF
--- a/src/utils/setup/save-asset-file.js
+++ b/src/utils/setup/save-asset-file.js
@@ -6,7 +6,7 @@ const { getMetadata } = require('../get-metadata');
 const metadata = getMetadata();
 const DB = metadata.database;
 
-const saveAssetFile = async asset => {
+const saveFile = async asset => {
     await fs.mkdir(path.join('static', path.dirname(asset.filename)), {
         recursive: true,
     });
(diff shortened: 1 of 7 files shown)


src/utils/setup/save-asset-file.js
```diff
@@ -26,17 +26,17 @@ const saveAssetFile = async asset => {
 };
 
 // Write all assets to static directory
-const saveAssetFiles = async (assets, stitchClient) => {
+const saveAssetFile = async (asset, stitchClient) => {
```
jestapinski, line 29: We hit stitch more often now here since we call this once per asset. This was actually faster than trying to memoize all of the assets in stitch and then filtering one out and writing out on demand.

greggb: That seems like it could be a lot. What's the delta in the number of calls to the network from our current build to this one?

jestapinski: It's going from one to ~252. I'm going to try to leverage the in-memory assets array we had before to try and move this back down to one

jestapinski:
Moving to WIP, I noticed after moving `saveAssetFiles` into `createPages` builds always are killed on my machine. I'm going to try to move how assets are pulled for more consistency.

src/utils/setup/save-asset-file.js
greggb: Thanks, let me know how that goes - I think making that many calls is asking for trouble (and could grow exponentially over time)

jestapinski:
Looks like the `fs` change also helped out here, prod builds seem to be back to normal. This is ready for review!

src/utils/setup/save-asset-file.js
greggb: We're back down to one stitch call now?

jestapinski: yes we are!

jestapinski:
Thanks! Apologies for the force push confusion

PR #305, opened 2020-04-10
Bump Deps

jestapinski:
This PR bumps a good portion of deps, including gatsby & plugins.

I added `{ Helmet }` to `tag.js` to remove non-default export warnings, and also updated lint-staged to remove extra `git add` commands at the end, since this is now done by default.

I didn't update prettier based on the `arg` to `(arg)` change, and also the leafy codeblock since they added a copy button column to all codeblocks that doesn't match our designs.


jestapinski:
Thanks!

PR #306, opened 2020-04-13
Extract "Create Page" logic and Test

jestapinski:
This PR simply moves the logic for creating an article page into another file and adding a few tests for it (things like properly adding series, related articles, etc.).


jestapinski:
Thanks!

PR #308, opened 2020-04-13
Use GraphQL to Query Pages to build

jestapinski:
Before, we were using an in-memory array of slugs to determine articles we want to build. This got a bit messy passing an array around, and with nodes in gatsby we can run a simple query instead for articles.

I wanted to get some feedback on using graphql here, since this is our first real foray into leveraging the data layer (outside of the site metadata stuff).

One nice advantage of moving queried elements into graphql is that we can access them in plugins (such as the rss plugin) or in other places without needing to pass global arrays around.


src/utils/setup/postprocess-document.js
```diff
@@ -1,25 +1,11 @@
...
 /*
   Modifies in-memory content based on what type of document (image, article)
 */
-const postprocessDocument = (
-    doc,
-    PAGE_ID_PREFIX,
-    pagesArray,
-    RESOLVED_REF_DOC_MAPPING
-) => {
+const postprocessDocument = (doc, PAGE_ID_PREFIX, RESOLVED_REF_DOC_MAPPING) => {
```
jestapinski, line 4: We might not even need this method anymore

jestapinski:
> I definitely think this is the right approach. This leads even further into creating a Stitch source plugin as well :)
> 
> Would like to get thoughts from @sophstad as well

For sure! There is one key note I want to make however. The content for each article is an AST, which has a shape that varies based on the article content (as far as I am aware). GraphQL likes a well-defined schema, so it would be impractical for us to try and store the content of each article in GraphQL, but we can do almost everything else.

sophstad:
This makes sense to me! Like you said the indetermine structure of these documents doesn't play nice with GraphQL, which is an unfortunate limitation. But I like this approach generally 💯

jestapinski:
Thanks all! This is now ready for a review

jestapinski:
Thanks!

PR #309, opened 2020-04-14
Add cypress testing for the home page

jestapinski:
This PR adds Cypress Integration testing to the home page. Right now, we can locally run these commands in Chrome, Edge, and Firefox after starting a development server.

I added checks for the following home page pieces:
- Featured Articles render (between 1 and 4) with images and refs to articles
- Learn MongoDB button links to learn page
- Twitch content renders with a thumbnail and pops open the content modal
- Events link has the expected href

We don't need specific data for these tests, since I just check the structure of components and props exist


jestapinski:
Thanks!

PR #310, opened 2020-04-14
Remove unused url arg from useEventData

jestapinski:
I noticed components using the `useEventData` hook were no longer using the url param supplied after refactoring events to use multiple sources. 

This PR simply removes references to that argument from caller components and the hook, because `fetchEventData` now has that url contained in its file.


jestapinski:
Thanks!

PR #311, opened 2020-04-14
Add Integration Testing for the Community Page

jestapinski:
This PR adds new integration testing for the `/community` page.

Since we might not want to continuously hit the events APIs, I was able to [mock them through cypress](https://docs.cypress.io/guides/guides/network-requests.html#Testing-Strategies). 

Interestingly enough, cypress can't mock `fetch`, so there are [some workarounds](https://github.com/cypress-io/cypress-example-recipes/tree/master/examples/stubbing-spying__window-fetch#readme) they recommend to do so (by deleting fetch on navigating to that page).

I added testing for the following pieces:
- mock a www.mongodb.com event API response and check the event object created is correct (with external link)
- mock a live.mongodb.com event API response and check the event object created is correct (with external link)
- check the "see all events" link references the correct page

### DIFF
--- a/cypress/support/commands.js
+++ b/cypress/support/commands.js
@@ -23,10 +23,6 @@ Cypress.Commands.add('mockEventsApi', () => {
 // we can fallback to XMLHttpRequests
 // https://github.com/cypress-io/cypress-example-recipes/tree/master/examples/stubbing-spying__window-fetch#readme
 
-Cypress.on('window:before:load', win => {
-    delete win.fetch;
-});
-
 Cypress.Commands.add('visitWithoutFetch', path => {
     cy.visit(path, {
         onBeforeLoad(win) {
--- a/cypress/support/commands.js
+++ b/cypress/support/commands.js
@@ -1,15 +1,3 @@
-// ***********************************************
-// This example commands.js shows you how to
-// create various custom commands and overwrite
-// existing commands.
-//
-// For more comprehensive examples of custom
-// commands please read more here:
-// https://on.cypress.io/custom-commands
-// ***********************************************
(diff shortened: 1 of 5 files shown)


cypress/support/commands.js
```diff
@@ -18,3 +6,31 @@ Cypress.Commands.add('closeModal', () => {
...
+
+// To stub requests with Cypress, we must remove fetch from the browser so
+// we can fallback to XMLHttpRequests
+// https://github.com/cypress-io/cypress-example-recipes/tree/master/examples/stubbing-spying__window-fetch#readme
+
+Cypress.on('window:before:load', win => {
+    delete win.fetch;
+});
+
+Cypress.Commands.add('visitWithoutFetch', path => {
```
jbouzi12, line 30: Thanks for the context/resources for mocking api calls. I like these helper functions as a means of doing this! When looking at this command as well as the `window:before:load` block I'm wondering if both of them are necessary. 

Does line 26 run on every cypress test and if so when would the `visitWithoutFetch` command be useful?

jestapinski: Great catch, yep these are redundant. I opted to keep the `visitWithoutFetch` so it is run less often. Thanks for that!

jbouzi12: No problem! I really like the `visitWithoutFetch` helper as it gives you more control as to whether or not you want to mock api calls 👍

jestapinski: Thanks! Yeah maybe we can use it on University as well somewhere

jestapinski:
Thanks Jensen!

PR #312, opened 2020-04-14
Add Integration Testing for Author Pages

jestapinski:
This PR adds some simple integration testing for the templated Author pages. Since writers can add articles or change their bios, I didn't want to test specific article/bio content, but just check it is still being rendered properly.

Once we have integration data from DOCSP then perhaps we could test against specific articles, etc., but I would rather add that in if stable then have to continuously update tests.

This PR adds testing for the following points:
- Checks some author image exists
- Checks the author's name is displayed
- Checks the author's bio is displayed
- Checks the article cards shown have all necessary pieces (made a cypress command for this to help with the learn page and other tag pages as well)


jestapinski:
Thanks!

PR #317, opened 2020-04-15
Fix codeblock border

jestapinski:
I thought I had added this into the codeblock dep update but it looks like I missed it.

This PR updates the border on the codeblock to be uniform throughout:

Screen Shot 2020-04-15 at 3 14 02 PM


jestapinski:
Thanks!

PR #319, opened 2020-04-15
Allow Authors to Exclude Learn Page Articles

jestapinski:
This PR enables the front-end to remove requested articles from being shown on the learn page. We are expecting this to come from the back-end as `learnPageExclude` as one of the page groups.

Specifically, if an article is requested to be removed from the learn page:
- We still create the article as normal
- We still show the article in the `tag` pages
- We do not show the article on the learn page
- We do not count the article in the filters on the learn page

I wrote a unit test for the exclusion, as well as locally checked a value for `learnPageExclude` to check out the behavior.

### DIFF
--- a/src/utils/setup/remove-excluded-articles.js
+++ b/src/utils/setup/remove-excluded-articles.js
@@ -1,0 +1,25 @@
+const removeExcludedArticles = (allArticles, excludedLearnPageArticles) => {
+    if (excludedLearnPageArticles && excludedLearnPageArticles.length) {
+        const filteredArticles = allArticles.filter(
+            article =>
+                !excludedLearnPageArticles.find(excludedArticle =>
+                    article.query_fields.slug.match(
+                        new RegExp(`^/?${excludedArticle}$`)
+                    )
+                )
+        );
+        // Warn writers if not all excludes were found
+        if (
+            allArticles.length - excludedLearnPageArticles.length <
+            filteredArticles.length
+        ) {
+            console.warn(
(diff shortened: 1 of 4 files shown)


src/utils/setup/remove-excluded-articles.js
```diff
@@ -0,0 +1,25 @@
...
+        const filteredArticles = allArticles.filter(
+            article =>
+                !excludedLearnPageArticles.find(excludedArticle =>
+                    article.query_fields.slug.match(
+                        new RegExp(`^/?${excludedArticle}$`)
+                    )
+                )
+        );
+        // Warn writers if not all excludes were found
+        if (
```
jestapinski, line 12: This seemed like a nice touch, but not sure if it is needed (or something we want to remove after this works for them).

src/utils/setup/remove-excluded-articles.js
```diff
@@ -0,0 +1,25 @@
+const removeExcludedArticles = (allArticles, excludedLearnPageArticles) => {
+    if (excludedLearnPageArticles && excludedLearnPageArticles.length) {
+        const filteredArticles = allArticles.filter(
+            article =>
+                !excludedLearnPageArticles.find(excludedArticle =>
+                    article.query_fields.slug.match(
+                        new RegExp(`^/?${excludedArticle}$`)
```
jestapinski, line 7: I would have loved a small optimization here by making RegExp objects in the outer loop, but we don't know if these slugs will have slashes or not, so the regex would be more complicated.

jestapinski:
Thanks!

PR #320, opened 2020-04-15
Add an RSS Feed for Articles

jestapinski:
This PR uses the [gatsby-plugin-feed](https://www.gatsbyjs.org/packages/gatsby-plugin-feed/) plugin to add an RSS feed for our articles at `/rss.xml` on production builds. GitHub won't let me attach an 'xml' file here, so I took a screenshot (and am happy to share the file on request).

For each article I included:
- Title
- Description
- Link
- Published Date

with results sorted so the newest articles are first. I'm not sure what else we would want to add in here.

Screen Shot 2020-04-15 at 5 19 31 PM

### DIFF
--- a/src/utils/setup/create-article-node.js
+++ b/src/utils/setup/create-article-node.js
@@ -17,6 +17,7 @@ const createArticleNode = (
         const content = {
             title: getNestedText(doc.query_fields['title']),
             description: getNestedText(doc.query_fields['meta-description']),
+            pubdate: doc.query_fields['pubdate'],
         };
         createNode({
             id: slug,
--- a/src/utils/setup/create-article-node.js
+++ b/src/utils/setup/create-article-node.js
@@ -10,7 +10,10 @@ const createArticleNode = (
 ) => {
     const filename = getNestedValue(['filename'], doc) || '';
     const isArticlePage =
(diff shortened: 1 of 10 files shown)


src/utils/setup/create-article-node.js
```diff
@@ -10,13 +10,17 @@ const createArticleNode = (
 ) => {
     const filename = getNestedValue(['filename'], doc) || '';
     const isArticlePage =
-        !filename.includes('images/') && filename.endsWith('.txt');
+        !filename.includes('images/') &&
+        filename.endsWith('.txt') &&
+        // Remove extraneous src/index.txt file in devhub-content repo
+        filename !== 'index.txt';
```
jestapinski, line 16: There is another piece of odd data in the content repo, `src/index.txt`. I asked Sophie and this isn't used as an article so we should exclude it. Its contents are empty.

greggb: This is good, but they should probably just remove this file rather than having us exclude extraneous files. I'll ask about this one.

greggb:
Checked in and we can just use the description for now which looks good in the preview 
image

jestapinski:
Excellent. My hunch is people may come back with more requests after it is out.

PR #321, opened 2020-04-16
Create github action for CI testing

jestapinski:
This PR adds a github action for CI testing a development build of the devhub with cypress to PRs to both staging and master.

**Test Duration (these tests are run in parallel)**
e2e tests run in ~2 minutes 30 seconds (biggest bottlenecks, starting gatsby server, npm installing)
unit tests run in ~1 minute 10 seconds (npm installing)

It looks like modules are cached between jobs (based on npm cache key values lining up), but each one has to `npm install` which can take ~30 seconds. These jobs are done in parallel, however, so it isn't a big cost in terms of overall duration. I also like having these jobs separate to keep workflows smaller and isolate failures.

Cache logs for e2e:
Screen Shot 2020-04-16 at 10 52 01 AM

Cache logs for unit testing:
Screen Shot 2020-04-16 at 10 51 39 AM

### DIFF
--- a/.github/workflows/devhub-ci-test.yaml
+++ b/.github/workflows/devhub-ci-test.yaml
@@ -12,6 +12,17 @@ jobs:
               uses: actions/setup-node@v1
               with:
                   node-version: '12.x'
+            - name: Cache node modules
+              uses: actions/cache@v1
+              with:
+                  path: node_modules
+                  key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+                  restore-keys: |
+                      ${{ runner.os }}-node-
+
+            - name: Install Dependencies
+              if: steps.cache.outputs.cache-hit != 'true'
+              run: npm ci
             - name: Create .env.development file
               run: |
                   touch .env.development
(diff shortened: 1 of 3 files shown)


.github/workflows/devhub-ci-test.yaml
```diff
@@ -0,0 +1,33 @@
...
+            - name: Use Node.js 12.x
+              uses: actions/setup-node@v1
+              with:
+                  node-version: '12.x'
+            - name: Create .env.development file
+              run: |
+                  touch .env.development
+                  echo "GATSBY_SITE=devhub" > .env.development; \
+                  echo "GATSBY_PARSER_USER=sophstad" >> .env.development; \
+                  echo "GATSBY_PARSER_BRANCH=master" >> .env.development; \
```
jestapinski, line 20: This can quickly be updated to new `env` values when DOCSP is ready.

greggb:
Would you mind trying out the cache package from this example? https://github.com/actions/cache/tree/4b0709a0d5bb7cd20f2ca9b9eb3dd620875faec9#example-workflow

Also trying `npm ci` instead of `npm install`? I' curious if we can get those install times down at all

greggb:
Hmm, looks like there's not much that can be done. Probably something like Yarn 2 (berry) would fix this, but not worth the effort :/

jestapinski:
Going to close this to stop spamming until I have the cache working, getting close on something.

PR #324, opened 2020-04-17
Add Cypress Github Action (With Caching)

jestapinski:
This PR adds a github action for CI testing a development build of the devhub with cypress to PRs to both staging and master. Furthermore, this PR adds conditional `npm install`ing only if a cached copy of `node_modules` is unavailable.

**Test Duration** (these tests are run in parallel)
e2e tests run in ~2 minutes
unit tests run in ~1 minute

I resolved issues with `npm installing` if a cache is available for `node_modules` which cut times down significantly on both test jobs. Restoring the cache still takes a bit of time, but this is likely because it is ~200MB compressed. Caching here is still faster than a fresh install.

### DIFF
--- a/.github/workflows/devhub-ci-test.yaml
+++ b/.github/workflows/devhub-ci-test.yaml
@@ -34,7 +34,7 @@ jobs:
                   echo "GATSBY_SITE=devhub" > .env.development; \
                   echo "GATSBY_PARSER_USER=sophstad" >> .env.development; \
                   echo "GATSBY_PARSER_BRANCH=master" >> .env.development; \
-                  echo "GATSBY_SNOOTY_DEV=true" >> .env.production; \
+                  echo "GATSBY_SNOOTY_DEV=true" >> .env.development; \
             - name: Cypress run
               uses: cypress-io/github-action@v1
               with:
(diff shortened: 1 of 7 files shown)


jestapinski:
Got the production build going for this action, but it also required a `fetch` polyfill (`whatwg-fetch`) to stub HTTP requests via Cypress. [This is a known issue](https://github.com/cypress-io/cypress/issues/95).

greggb:
> Got the production build going for this action, but it also required a `fetch` polyfill (`whatwg-fetch`) to stub HTTP requests via Cypress. [This is a known issue](https://github.com/cypress-io/cypress/issues/95).

Seems like it also brought the install time back up to where it was before the caching changes

jestapinski:
> > Got the production build going for this action, but it also required a `fetch` polyfill (`whatwg-fetch`) to stub HTTP requests via Cypress. [This is a known issue](https://github.com/cypress-io/cypress/issues/95).
> 
> Seems like it also brought the install time back up to where it was before the caching changes

This is because the cache didn't match this time! Since we added a new dep the node modules cache was no longer valid, so we had to install anyway.

jestapinski:
^ that was it, the build as back closer to 2 minutes with a valid cache. Still a touch longer because of the build

jestapinski:
Thanks for the pointers!

PR #325, opened 2020-04-17
Generalize Twitch Video Logic

tahiya-mongo:
Pulled out just the logic for accessing videos from `useTwitchApi.js` to reuse it for the video/media page.

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -152,3 +152,11 @@ export const devhubMapping = {
     nodejs: 'Node.js',
     java: 'Java',
 };
+
+// Twitch API constants
+export const TWITCH_API_ENDPOINT = 'https://api.twitch.tv/helix/';
+export const TWITCH_CLIENT_ID = '041r2glmgub2pt357ss0la44j2sz95';
+export const TWITCH_MDB_CHANNEL_ID = '467752938';
+export const TWITCH_headers = { 'Client-ID': TWITCH_CLIENT_ID };
+export const TWITCH_STREAMS_URL = `${TWITCH_API_ENDPOINT}streams?user_id=${TWITCH_MDB_CHANNEL_ID}`;
+export const TWITCH_VIDEO_URL = `${TWITCH_API_ENDPOINT}videos?user_id=${TWITCH_MDB_CHANNEL_ID}&first=`;
--- a/src/constants.js
+++ b/src/constants.js
@@ -157,6 +157,6 @@ export const devhubMapping = {
 export const TWITCH_API_ENDPOINT = 'https://api.twitch.tv/helix/';
 export const TWITCH_CLIENT_ID = '041r2glmgub2pt357ss0la44j2sz95';
 export const TWITCH_MDB_CHANNEL_ID = '467752938';
-export const TWITCH_headers = { 'Client-ID': TWITCH_CLIENT_ID };
+export const TWITCH_HEADERS = { 'Client-ID': TWITCH_CLIENT_ID };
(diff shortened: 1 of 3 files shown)


src/utils/fetch-twitch-videos.js
```diff
@@ -0,0 +1,22 @@
+import { get } from './request';
+
+const TWITCH_API_ENDPOINT = 'https://api.twitch.tv/helix/';
+const CLIENT_ID = '041r2glmgub2pt357ss0la44j2sz95';
+const MDB_CHANNEL_ID = '467752938';
+const video_URL = `${TWITCH_API_ENDPOINT}videos?user_id=${MDB_CHANNEL_ID}&first=`;
+const headers = { 'Client-ID': CLIENT_ID };
+
+// Fetches videos from twitch api
+const fetchTwitchVideos = async (videoLimit = 1) => {
```
jestapinski, line 10: It would be great to use this in the `use-twitch-api` hook to abstract out some small details (such as having to get `.data` on the response, etc.). Perhaps in the `use-twitch-api` hook we can do something like the following with this:

```
            if (!currentStream && videoLimit) {
                const videoResp = await fetchTwitchVideos(some video limit);
                // return the whole array of videos, but ignore pagination for now
                setVideos(videoResp);
            }
```

src/utils/fetch-twitch-videos.js
```diff
@@ -0,0 +1,22 @@
+import { get } from './request';
+
+const TWITCH_API_ENDPOINT = 'https://api.twitch.tv/helix/';
```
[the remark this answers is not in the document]
jestapinski: Actually another thought, we might want these in another file altogether since the hooks use them and places that will use `fetchTwitchData` will also need them. Perhaps another constants file makes sense, what do you think?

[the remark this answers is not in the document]
tahiya-mongo: I was actually just thinking about that. It definitely makes more sense to have them in a separate file

src/constants.js
```diff
@@ -152,3 +152,11 @@ export const devhubMapping = {
     nodejs: 'Node.js',
     java: 'Java',
 };
+
+// Twitch API constants
+export const TWITCH_API_ENDPOINT = 'https://api.twitch.tv/helix/';
+export const TWITCH_CLIENT_ID = '041r2glmgub2pt357ss0la44j2sz95';
+export const TWITCH_MDB_CHANNEL_ID = '467752938';
+export const TWITCH_headers = { 'Client-ID': TWITCH_CLIENT_ID };
```
jestapinski, line 160:
```suggestion
export const TWITCH_HEADERS = { 'Client-ID': TWITCH_CLIENT_ID };
```

small nit: This should also be in caps

src/utils/use-twitch-api.js
```diff
@@ -1,13 +1,11 @@
...
-const MDB_CHANNEL_ID = '467752938';
-const STREAMS_URL = `${TWITCH_API_ENDPOINT}streams?user_id=${MDB_CHANNEL_ID}`;
-const video_URL = `${TWITCH_API_ENDPOINT}videos?user_id=${MDB_CHANNEL_ID}&first=`;
-
-const headers = { 'Client-ID': CLIENT_ID };
+import {
+    TWITCH_headers,
+    TWITCH_STREAMS_URL,
+    TWITCH_VIDEO_URL,
+} from '../constants';
```
jestapinski, line 8: Nice! Yeah I think having these constants more easily accessible from both use cases in a separate file makes sense.

PR #326, opened 2020-04-17
Mobile Article Layout Issues

jestapinski:
This PR fixes two things for mobile articles:
- Adds an overflow to the table directive component to allow tables to be scrollable horizontally (This is broken on articles such as [this one](https://developer.mongodb.com/article/map-terms-concepts-sql-mongodb) in prod)
- Fixes the `useMedia` hook during static rendering. While building, the hook value is the default value, and then gets re-calculated when served. This fixes an issue with the top nav where refreshing on an article page would look like this:
Screen Shot 2020-04-17 at 3 50 57 PM

Not 100% sure why this works, but the component is now properly re-rendering with the updated hook value, so this seems to have fixed a logical issue in the `useEffect` of `useMedia`.

### DIFF
--- a/src/components/ListTable.js
+++ b/src/components/ListTable.js
@@ -20,6 +20,10 @@ const Table = styled('table')`
     }
 `;
 
+const TableContainer = styled('div')`
+    overflow-x: auto;
+`;
+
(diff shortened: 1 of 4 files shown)


jestapinski:
Thanks!

PR #330, opened 2020-04-20
Address Lighthouse/Screen Reader Accessibility Issues

jestapinski:
I audited different types of pages on DevHub using the Lighthouse dev tool as well as went through portions of the site with a screen reader and found some issues to address to better support accessibility tools. Specifically, this PR:

- Adds `aria-labels` to icons used on the footer or in article sharing
- Adds `aria-label="play"` to the play button
- Adds `aria-label="close"` to the modal close button
- Removes some `id` fields in icons (Lighthouse found some duplicate `id` errors)
- Adds `lang="en"` to the top level `html` tag using helmet
- Adds `option` role to dropdown options in the `Select` component
- Adds proper `title` for the tag pages


src/components/dev-hub/global-nav.js
```diff
@@ -74,7 +74,7 @@ export default () => {
     return (
         <GlobalNav>
             <NavContent>
-                <HomeLink to="/">
+                <HomeLink aria-label="Home" to="/">
```
greggb, line 77: Would title work here as well, or is aria-label the recommended approach?

jestapinski: Specifically for screen-readers, reading a `title` [isn't always enabled by default](https://www.mediacurrent.com/blog/dont-rely-title-attribute-accessibility-seo/):

```
A person who uses a screen reader because they are blind or have low vision will likely never have the title attribute read to them. Some screen readers don’t support the title attribute and those that do, don’t read the title attribute by default. It requires changing a setting for the title to be read, which not all users do.
```

Are there other use cases for the `title` attribute?

greggb: Wow, TIL. I was always under the impression title served two purposes
- add a hover tooltip
- add accessibility text

jestapinski: `title` does add that tooltip, so it seems the most bulletproof way would be to have both an `aria-label` as well as a `title`, but it also seems redundant. Thoughts?

greggb: I'm fine with just the label

jestapinski:
Thanks!

PR #332, opened 2020-04-20
Create method to pull Libsyn Podcasts

tahiya-mongo:
This PR pulls down mongodb podcast information hosted at https://mongodb.libsyn.com/rss and parses it to get an array of podcasts.


src/utils/get-lybsin-podcasts.js
```diff
@@ -0,0 +1,47 @@
+// Fetches podcast info from https://mongodb.libsyn.com/rss
+
+const RSS_URL = `https://mongodb.libsyn.com/rss`;
+let prodcastList = [];
```
jestapinski, line 4:
```suggestion
const prodcastList = [];
```
nit, but we never re-assign `podcastList` so we can use `const` here. This also can get a bit confusing since there is a `podcastList` in local scope for `getLybsinPodcasts`. Perhaps this variable can just move into `parsePodcasts`?

src/utils/get-lybsin-podcasts.js
```diff
@@ -0,0 +1,47 @@
...
+
+export const parsePodcasts = podcastXML => {
```
jestapinski, line 6: I think this logic would be great to write a unit test for! Any kind of parsing in general it's great to capture that logic in a test to represent behavior, what do you think?

src/utils/get-lybsin-podcasts.js
```diff
@@ -0,0 +1,47 @@
...
+    for (var podcast of podcastXML) {
```
jestapinski, line 7: This is another great use case for the JS `map` function, since we want to take an array of XML podcasts and "map" it to JSON.  [Check this out](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map)!

src/utils/get-lybsin-podcasts.js
```diff
@@ -0,0 +1,51 @@
...
+
+    var options = {
+        ignoreAttributes: false,
+        parseAttributeValue: true,
+    };
+
+    try {
+        var jsonObj = parser.parse(podcastXML, options, true);
+        const podcasts = jsonObj.rss.channel.item;
+        podcasts.pop(); //removing the introductory podcast
```
jestapinski, line 26: Is there a specific reason we want to remove the intro podcast?

src/utils/get-lybsin-podcasts.js
```diff
@@ -0,0 +1,51 @@
...
+    };
+    return podcastJSON;
+};
+
+export const parsePodcasts = podcastXML => {
+    var parser = require('fast-xml-parser');
+
+    var options = {
+        ignoreAttributes: false,
+        parseAttributeValue: true,
```
jestapinski, line 20: We can also set the `attributeNamePrefix` to '' to get rid of that odd `@_` for url.

src/utils/get-lybsin-podcasts.js
```diff
@@ -0,0 +1,51 @@
+// Fetches podcast info from https://mongodb.libsyn.com/rss
+
+const RSS_URL = `https://mongodb.libsyn.com/rss`;
+
+export const simplifyPodcast = podcast => {
+    const podcastJSON = {
```
jestapinski, line 6: We also will likely need the image as well here to show!

src/utils/get-lybsin-podcasts.js
```diff
@@ -0,0 +1,51 @@
...
+        title: podcast['title'],
+        publishDate: podcast['pubDate'],
+        summary: podcast['itunes:summary'],
+        url: podcast['enclosure']['@_url'],
+    };
+    return podcastJSON;
+};
+
+export const parsePodcasts = podcastXML => {
+    var parser = require('fast-xml-parser');
```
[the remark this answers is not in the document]
tahiya-mongo: Sorry, I forgot to commit the updated `package.json` file

src/utils/get-lybsin-podcasts.js
```diff
@@ -0,0 +1,51 @@
...
+
+    var options = {
+        ignoreAttributes: false,
+        parseAttributeValue: true,
+    };
+
+    try {
+        var jsonObj = parser.parse(podcastXML, options, true);
+        const podcasts = jsonObj.rss.channel.item;
```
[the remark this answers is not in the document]
tahiya-mongo: cool stuff!

src/utils/get-lybsin-podcasts.js
```diff
@@ -0,0 +1,51 @@
+// Fetches podcast info from https://mongodb.libsyn.com/rss
```
jestapinski, line 1: Did we want to remove this file at this time?

src/utils/fetch-lybsin-podcasts.js
```diff
@@ -0,0 +1,53 @@
...
+    return [];
+};
+
+const getLybsinPodcasts = async () => {
+    try {
+        const response = await fetch(RSS_URL);
+        if (response) {
+            const podcastXML = await response.text();
+            const podcastList = parsePodcasts(podcastXML);
+            console.log(podcastList);
```
jestapinski, line 44:
```suggestion
```

src/utils/fetch-lybsin-podcasts.js
```diff
@@ -0,0 +1,53 @@
...
+        const parsedPodcasts = podcasts.map(simplifyPodcast);
+        return parsedPodcasts;
+    } catch (error) {
+        console.log(error.message);
+    }
+
+    return [];
+};
+
+const getLybsinPodcasts = async () => {
```
jestapinski, line 38:
```suggestion
const fetchLybsinPodcasts = async () => {
```

Let's have the name match the filename here

src/utils/fetch-lybsin-podcasts.js
```diff
@@ -0,0 +1,53 @@
+import parser from 'fast-xml-parser';
+import dlv from 'dlv';
+
+// Fetches and parses podcast info from https://mongodb.libsyn.com/rss
+
+const RSS_URL = `https://mongodb.libsyn.com/rss`;
+
+export const simplifyPodcast = podcast => {
```
jestapinski, line 8: Are we `export`ing these two other functions for any reason?

tahiya-mongo: I was only exporting them because I was following the `fetch-event-data.js` file. Just removed them since we're won't  be calling them from anywhere else.

PR #334, opened 2020-04-21
Remove unused constants and utils, move twitch hook to hooks

jestapinski:
We noticed some unused files/constants in the repo that seemed snooty-specific so this PR removes them.

I also move the `use-twitch-api` hook into `/hooks`.


jestapinski:
Thanks all!

PR #336, opened 2020-04-22
Switch parsed data source to CI repo for CI testing

jestapinski:
I recently got access to use the snooty parser and store parsed data on Atlas from the DOCSP team, so this PR switches our testing data from using Sophie's parsed data to my integration testing content repo.

This allows us to keep data consistent across CI runs, and also test new cases by customizing our integration testing data to fit our needs (we now can control our own content repo, parse it, test against it, and see it in staging before deployment).

### DIFF
--- a/.github/workflows/devhub-ci-test.yaml
+++ b/.github/workflows/devhub-ci-test.yaml
@@ -32,7 +32,7 @@ jobs:
               run: |
                   touch .env.production
                   echo "GATSBY_SITE=devhub" > .env.production; \
-                  echo "GATSBY_PARSER_USER=sophstad" >> .env.production; \
+                  echo "GATSBY_PARSER_USER=jordanstapinski" >> .env.production; \
                   echo "GATSBY_PARSER_BRANCH=master" >> .env.production; \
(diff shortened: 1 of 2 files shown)


.github/workflows/devhub-ci-test.yaml
```diff
@@ -32,7 +32,7 @@ jobs:
               run: |
                   touch .env.production
                   echo "GATSBY_SITE=devhub" > .env.production; \
-                  echo "GATSBY_PARSER_USER=sophstad" >> .env.production; \
+                  echo "GATSBY_PARSER_USER=jordanstapinski" >> .env.production; \
```
greggb, line 35: could we have a devhub user? or is it a personal account only?

jestapinski: I can look into this, it by default uses the account name of your Mac (`echo $USER`). It may be difficult to change. This is how the data is also represented on Atlas

If my Mac had the accountname `devhub` than all would be well :)

greggb: if only :)

jestapinski:
Thanks!

PR #338, opened 2020-04-23
Add image, other meta to rss feed

jestapinski:
This PR uses our favicon for the RSS feed, and also adds `site_url` and `feed_url` which are passed to the rss `feedOptions` [here](https://www.npmjs.com/package/rss#itemoptions) through Gatsby.

I also pulled this object out of the config files and just reference it in config and config-prod instead.

This adds the following node to the RSS XML:
Screen Shot 2020-04-23 at 10 46 47 AM

Not sure the best way to preview this, but it seems this is the correct field for `rss` to have the favicon.

### DIFF
--- a/src/utils/setup/article-rss-feed.js
+++ b/src/utils/setup/article-rss-feed.js
@@ -1,0 +1,14 @@
+const { rssFeedArticleData } = require('../../queries/rss-feed-article-data');
+const { serializeRssData } = require('./serialize-rss-data');
+
+const articleRssFeed = {
+    serialize: serializeRssData,
+    query: rssFeedArticleData,
+    output: '/rss.xml',
+    title: 'MongoDB Developer Hub RSS Feed',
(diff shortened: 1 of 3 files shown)


src/utils/setup/article-rss-feed.js
```diff
@@ -0,0 +1,14 @@
+const { rssFeedArticleData } = require('../../queries/rss-feed-article-data');
+const { serializeRssData } = require('./serialize-rss-data');
+
+const articleRssFeed = {
+    serialize: serializeRssData,
+    query: rssFeedArticleData,
+    output: '/rss.xml',
+    title: 'MongoDB Developer Hub RSS Feed',
+    image_url: 'https://media.mongodb.org/favicon.ico',
```
greggb, line 9: This might actually need to be a bigger image for some readers, but it'll work for now. Couldn't find a good reference for sizes on this online

jestapinski:
Thanks!

PR #341, opened 2020-04-24
Add cypress testing for article pages

jestapinski:
This PR adds cypress testing for a specific article off of the integration content repo (`/article/3-things-to-know-switch-from-sql-mongodb/`).

I added the following general tests:
- Tested the header is populated with all of the important information
- Tested the contents menu brings the user to a specific header when requested
- Social URLs to LinkedIn, Twitter, Facebook work as expected
- A series is shown with links populated
- Proper SEO/meta tags are maintained (checking various fields on `meta` to ensure og/meta fields are populated as we expect)

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -1,4 +1,4 @@
 {
-    "baseUrl": "http://localhost:8000",
+    "baseUrl": "http://localhost:9000/master/devhub/jordanstapinski/HEAD",
     "viewportWidth": 1280
 }
--- a/cypress.json
+++ b/cypress.json
@@ -1,4 +1,4 @@
 {
-    "baseUrl": "http://localhost:9000/master/devhub/jordanstapinski/HEAD",
+    "baseUrl": "http://localhost:8000",
     "viewportWidth": 1280
 }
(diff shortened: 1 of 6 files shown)


cypress.json
```diff
@@ -1,4 +1,4 @@
 {
-    "baseUrl": "http://localhost:9000/master/devhub/jordanstapinski/HEAD",
+    "baseUrl": "http://localhost:8000",
```
jestapinski, line 2: This is a local setting, so I removed the prefix path from it so local development can test without having to use prefix paths (we use prefix paths for the GH Actions test)

cypress/integration/article.js
```diff
@@ -0,0 +1,124 @@
...
+            cy.get('a').should('have.length', 3);
+            // Check other one goes to article
+            cy.get('a')
+                .first()
+                .should('have.prop', 'href')
+                .and('include', '/article/map-terms-concepts-sql-mongodb');
+        });
+    });
+
+    it('should have proper SEO, open graph, and twitter tags', () => {
```
jestapinski, line 92: This felt like a weird Cypress test, but we had no SEO testing so I believed this was a good way to add consistency in.

Perhaps a snapshot test on the SEO would be more philosophical, but using our integration data here gives a good test as well and runs through the parser.

greggb: I think this is reasonable. You're right, my first thought is we should test the SEO component itself, but if we're trying to confirm the page contents this makes sense as an 'article e2e test'

jestapinski:
Thanks!

PR #342, opened 2020-04-24
Reorganize Series Logic for Past/Upcoming Series Items

jestapinski:
This PR lays some logical groundwork for an upcoming change to the series component. We are going to want to style links a bit differently whether its article is before or after the current one in the series (to promote continuing the series).

This PR changes logic to pass the active item index in the series to the series component instead of the title, so it will be easier to get past or upcoming series items by comparing indices.

I also moved conditional styling into the css prop.

Upcoming design (not in this PR):
Screen Shot 2020-04-24 at 2 40 58 PM

### DIFF
--- a/src/components/dev-hub/article-series.js
+++ b/src/components/dev-hub/article-series.js
@@ -1,0 +1,41 @@
+import React from 'react';
+import dlv from 'dlv';
+import Series from './series';
+
+const ArticleSeries = ({ allSeriesForArticle, slugTitleMapping, title }) => {
+    // Handle if this article is not in a series or no series are defined
+    if (!allSeriesForArticle) return null;
+    const getMappedSeries = seriesSlugs => {
+        if (!seriesSlugs || !seriesSlugs.length) return null;
+
+        let hasSeenActiveSlug = false;
+        const getSeriesArticlePosition = articleTitle => {
+            // Find if an article is the current one in the series, or is upcoming/past
+            if (hasSeenActiveSlug) {
+                return 'upcoming';
+            } else if (articleTitle === title) {
+                hasSeenActiveSlug = true;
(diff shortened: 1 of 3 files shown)


src/templates/article.js
```diff
@@ -89,14 +89,20 @@ const ArticleSeries = ({ allSeriesForArticle, slugTitleMapping, title }) => {
...
-        <>
-            <Series name={series} currentStep={title}>
-                {getMappedSeries(allSeriesForArticle[series])}
-            </Series>
-            <br />
-        </>
-    ));
+    return Object.keys(allSeriesForArticle).map(series => {
+        const seriesItems = getMappedSeries(allSeriesForArticle[series]);
+        const activeSeriesIndex = seriesItems.findIndex(
```
jbouzi12, line 94: I wonder if we can reduce the number of loops we have here now that it's 3. What do you think?

jestapinski: Oh yeah great point, we can definitely move this check into the `getMappedSeries` helper, nice find!

src/components/dev-hub/article-series.js
```diff
@@ -0,0 +1,41 @@
...
+            } else if (articleTitle === title) {
+                hasSeenActiveSlug = true;
+                return 'active';
+            }
+            return 'past';
+        };
+
+        const mappedSeries = seriesSlugs.map(slug => {
+            const title = dlv(slugTitleMapping, [slug, 0, 'value'], slug);
+            const position = getSeriesArticlePosition(title);
```
jestapinski, line 25: Here is where I addressed the loop issue, I also pulled this component out because it was getting a bit long to leave in `article.js`.

jestapinski:
Thanks all!

PR #345, opened 2020-04-27
Cypress Tests for Tag Pages

jestapinski:
This PR adds some simple cypress tests for the tag pages. We test:
- The header and breadcrumbs are populated with the relevant tag
- Articles are being populated on the tag page
- An article that has the tag `SQL` is displayed on the `SQL` tag page
- The `robots` meta name exists with the value `noindex`


jestapinski:
Thanks Jensen!

PR #347, opened 2020-04-29
Fix Chrome Play Button Flashing Bug

jestapinski:
This PR fixes an odd chrome-only bug. When on the index page and hovering over the play button in the twitch video modal, the background would flash white momentarily.

Adding a background to the `GlobalWrapper` seemed to address the problem.

Before:
Screen Shot 2020-04-29 at 11 32 35 AM

After:
Screen Shot 2020-04-29 at 11 34 21 AM


jestapinski:
Thanks all!

PR #348, opened 2020-04-29
Fix Codeblock Border on Safari

jestapinski:
This PR fixes an issue only on safari where the codeblock line number border was on all four sides of the `div` (and wider than it should have been). It should only be on the right side. Explicitly setting border widths to 0 fixed this.

Before:
Screen Shot 2020-04-29 at 1 44 48 PM

After:
Screen Shot 2020-04-29 at 1 44 55 PM

### DIFF
--- a/src/components/dev-hub/codeblock.js
+++ b/src/components/dev-hub/codeblock.js
@@ -36,12 +36,12 @@ const StyledCode = styled(Code)`
     padding-right: ${size.xlarge};
     padding-top: ${size.large};
     ${({ numdigits }) =>
-        `padding-left: calc(${LEAFY_LINENUMBER_PADDING}px + ${numdigits *
-            size.stripUnit(size.xsmall)}px)`};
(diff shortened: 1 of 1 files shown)


jestapinski:
> Lgtm and looks like it fixed the issue. What caused the issue here?

It looks like Safari was reading `border: none` which cleared existing borders, but then adding the `border-image` reset this so other borders were being displayed as well. This change makes sure the order is right by reading the border image and then width for all sides (not just the right!)

jbouzi12:
> > Lgtm and looks like it fixed the issue. What caused the issue here?
> 
> It looks like Safari was reading `border: none` which cleared existing borders, but then adding the `border-image` reset this so other borders were being displayed as well. This change makes sure the order is right by reading the border image and then width for all sides (not just the right!)

Ah yea that will do it. Looks like the re-order did the trick!

jestapinski:
Thanks!

PR #259, opened 2020-03-23
WIP Jb/update storybook config

jbouzi12:
This branch aims to update our configuration to ensure proper parsing of styled components in `mdx` files and allow proper linting of `mdx` files through `prettier`. 

### Current State of branch
- styled components seem to be rendering correctly, although more testing needs to be done:
    - Prior to adding `@emotion/babel-preset-css-prop` none of the `Card` stories would render, but they are now rendering properly.
    - Other stories that previously had issues to test creating: `tooltip/content menus`, `breadcrumbs`
- `prettier` throws linting errors when comitting. This still needs to be addressed.
### Still Needs
- fixes to ensure `prettier` properly parses `mdx1

PR #349, opened 2020-04-29
Update and Fix Storybook Configuration

jestapinski:
This PR builds off of Jensen's work in #259 and finishes adding in fonts and other styles to have components render as expected and finish setting up storybook for devhub. There should not be any changes to the actual devhub site.

To test:
- Run `yarn storybook` (yep yarn works)
- Check out the storybook!

The changes I made after #259 to get this going were:
- Took style declarations in `layout.js` and applied them to `.storybook/preview-head.html` so they would be picked up by each component. `preview-head.html` is evaluated before each component is rendered, so our "global" styles will be picked up by each story component.
- Added `import 'typeface-fira-mono'` to the storybook `theme.js` file so our proper typeface is picked up (this also happens in `layout.js`), a different Fira Mono was being used for some reason.
- Cleaned up some broken stories (charts id was out of date, series was out of date)

### DIFF
--- a/src/components/dev-hub/stories/card.stories.mdx
+++ b/src/components/dev-hub/stories/card.stories.mdx
@@ -1,0 +1,87 @@
+import { Meta, Story, Preview } from '@storybook/addon-docs/blocks';
+import Card from '../card';
+import { colorMap } from '../theme';
+import mockCardImage from '../../../images/360-mock-img.png';
+import { mapTagTypeToUrl } from '../../../utils/map-tag-type-to-url';
+
+<Meta title="Cards" />
+
+# Cards
+
+A card can have a link, image, title, and description.
+<Preview>
+    <Story name="default card">
+        <div>
+            <Card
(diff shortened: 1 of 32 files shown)


jestapinski:
Interesting note, I had not run into issues when committing stories. Do we know if something specific could not be committed?

src/components/dev-hub/stories/card.stories.mdx
```diff
@@ -0,0 +1,87 @@
...
+    <Story name="default card">
+        <div>
+            <Card
+                distinct
+                image={mockCardImage}
+                maxTitleLines={3}
+                title="Title for Card"
+                description="Description for card that has a link."
+                href="#cards"
+            ></Card>
```
greggb, line 22: Think we can use a self closing tag here `/>` to fix the syntax highlighting

jestapinski:
Thanks!

PR #351, opened 2020-04-30
Fix default props in card-list, update CardContainer

jestapinski:
This PR does a few things with the `CardList`:

1. Fixes an issue with default props where instead of using default values, we null check and use `[]` if passed `null` (we were being passed `null` which overrode the previous default).
2. Moves `<CardContainer>` in `CardList` to wrap all the cards
3. Adds commentary about future work to come with this component to handle different types of items and articles.

Pages to check:
- /learn
- /tag
- /media

### DIFF
--- a/src/pages/media.js
+++ b/src/pages/media.js
@@ -16,7 +16,7 @@ export default () => {
     return (
         <Layout>
             {(isLoading || isLoadingPodcasts) && <P>Loading...</P>}
-            {videos && <CardList videos={videos} podcasts={podcasts} />}
+            <CardList videos={videos} podcasts={podcasts} />
             {(error || errorPodcasts) && (
                 <P>Check back later for upcoming media contents</P>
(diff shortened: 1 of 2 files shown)


src/pages/media.js
```diff
@@ -16,7 +16,7 @@ export default () => {
     return (
         <Layout>
             {(isLoading || isLoadingPodcasts) && <P>Loading...</P>}
-            {videos && <CardList videos={videos} podcasts={podcasts} />}
+            <CardList videos={videos} podcasts={podcasts} />
```
jestapinski, line 19: This line was part of the problem, if we had videos first but not podcasts, podcasts would pass a value `null` which would cause issues in the cardlist. I moved this type of checking to the cardlist so callsites don't need to worry about input validation.

jestapinski:
Thanks!

PR #355, opened 2020-05-01
Tab component for the Learn page

tahiya-mongo:
This PR includes the `Tab` component for the `learn.js` page. The tab component will allow us to toggle between different media types and articles.


tahiya-mongo:
@jestapinski The component isn't very responsive to different screen sizes as is. Any feedback regarding that would be greatly appreciated.

src/components/dev-hub/tab.js
```diff
@@ -0,0 +1,116 @@
+import { React, useState } from 'react';
+import styled from '@emotion/styled';
+import { colorMap, size } from './theme';
+import { P } from './text';
+import CardList from './card-list';
+import useAllVideos from '../../hooks/use-all-videos';
+import usePodcasts from '../../hooks/use-podcasts';
+
+const Tab = styled('div')`
```
jestapinski, line 9: We should let a component using the `Tab` define these values as needed (through using `styled` or something) like the media page, so these will likely move out of this component.

This lets us tweak the tab to fit a specific page's liking much more easily and also remove some code from the component itself.

src/components/dev-hub/tab.js
```diff
@@ -0,0 +1,116 @@
...
+    margin-bottom: 30px;
+`;
+
+const TabButton = styled('button')`
+    float: right;
+    border: none;
+    outline: none;
+    cursor: pointer;
+    transition: 0.3s;
+    font-size: 18px;
```
jestapinski, line 24: Can we use one of the `fontSize` values in `theme.js` (with its `line-height`?)? (or use one of the `P` flavors below?)

src/components/dev-hub/tab.js
```diff
@@ -0,0 +1,116 @@
...
+    &:hover {
+        color: white;
+    }
+    &:focus {
+        border-bottom-color: ${colorMap.darkGreen};
+    }
+`;
+
+const AllContentStyle = {
+    marginRight: '818px',
```
jestapinski, line 43: This is where the responsive issues are coming from! What happens is sometimes we run into a situation where this space isn't available and we get wonky behavior, but the browser is correct in thinking it needs to put this `818px` away.

We'll talk about how we can remedy this in the check in, but let's use something like [flexbox](https://css-tricks.com/snippets/css/a-guide-to-flexbox/) or css-grid to provide a more responsive layout!

src/components/dev-hub/tab.js
```diff
@@ -0,0 +1,116 @@
...
+        setShowVideos(false);
+        setShowPodcasts(false);
+    };
+
+    return (
+        <div>
+            <Tab collapse>
+                <TabButton onClick={() => displayVideos()}>Videos</TabButton>
+                <TabButton onClick={() => displayPodcasts()}>
+                    Podcasts
```
jestapinski, line 93: What would happen if we tried using one of the `P` flavors from `text.js` here? That would let us remove some of the css above.

src/components/dev-hub/tab.js
```diff
@@ -0,0 +1,116 @@
...
+        error: errorPodcasts,
+        isLoading: isLoadingPodcasts,
+    } = usePodcasts();
+
+    const [showVideos, setShowVideos] = useState(false);
+    const [showPodcasts, setShowPodcasts] = useState(false);
+    const [showArticles, setShowArticles] = useState(false);
+    const [showAll, setShowAll] = useState(true);
+
+    const displayVideos = () => {
```
jestapinski, line 60: I like how we used state hooks to manage which to show! Using four state variables seems a bit high, could we get away with one variable that is something like `'all'` or `'articles'`, etc.? That could remove the need for these four helpers

src/components/dev-hub/tab.js
```diff
@@ -0,0 +1,116 @@
...
+    const displayAll = () => {
+        setShowAll(true);
+        setShowArticles(false);
+        setShowVideos(false);
+        setShowPodcasts(false);
+    };
+
+    return (
+        <div>
+            <Tab collapse>
```
jestapinski, line 90:
```suggestion
            <Tab>
```
I don't believe we are using this prop?

src/components/dev-hub/tab.js
tahiya-mongo: Using styled `P` tag seems to be getting in the way of `:hover`. I left the styles as they are for now. Let me know if there is a way to work around it.

src/components/dev-hub/tab.js
```diff
@@ -0,0 +1,79 @@
...
+    border-bottom: 3px solid transparent;
+
+    font-size: ${fontSize.default};
+    color: ${colorMap.greyDarkOne};
+    font-family: 'Fira Mono';
+
+    ${({ isActive }) => (isActive ? activeStyles : defaultStyles)}
+`;
+
+const LeftContent = styled('div')``;
```
jestapinski, line 43: Since we aren't actually using styles for these `div`s we can just use `div` tags here :)

src/components/dev-hub/tab.js
```diff
@@ -0,0 +1,79 @@
...
+
+const RightContent = styled('div')``;
+
+export default ({ handleClick, tabTexts, activeItem }) => {
+    return (
+        <Tab>
+            <RightContent>
+                {tabTexts.slice(0, 1).map(item => {
```
jestapinski, line 51: Instead of slicing, lets replace `tabTexts` with two props `leftTabs` and `rightTabs` to make this component more flexible

src/components/dev-hub/tab.js
```diff
@@ -0,0 +1,79 @@
...
+                    const isActive = item === activeItem ? 'true' : null;
```
jestapinski, line 52:
```suggestion
                    const isActive = item === activeItem;
```
this gives us a boolean we can get the same effect with!

src/components/dev-hub/tab.js
```diff
@@ -0,0 +1,79 @@
...
+                    return (
+                        <TabButton
+                            isActive={isActive}
```
jestapinski, line 55: Since we are mapping we will need to add a `key` prop to these `TabButtons` (and the ones on the right as well)

src/components/dev-hub/tab.js
```diff
@@ -0,0 +1,79 @@
+import React from 'react';
+import styled from '@emotion/styled';
+import { css } from '@emotion/core';
+import { colorMap, size, fontSize } from './theme';
+
+const Tab = styled('div')`
+    display: flex;
+    justify-content: space-between;
```
jestapinski, line 8: Nice!

src/pages/media.js
```diff
@@ -13,10 +14,34 @@ export default () => {
         isLoading: isLoadingPodcasts,
     } = usePodcasts();
 
+    const [showContent, setShowContent] = useState('All');
+    const [activeItem, setActiveItem] = useState('All');
```
jestapinski, line 18: Having two state vars for this is redundant, can we just have one? `[activeItem, setActiveItem]` and use that?

src/pages/media.js
```diff
@@ -13,10 +14,34 @@ export default () => {
...
-            <CardList videos={videos} podcasts={podcasts} />
+
+            {/*TODO: display articles once we have access to them*/}
+
+            {showContent === 'All' && (
+                <CardList videos={videos} podcasts={podcasts} />
+            )}
+            {showContent === 'Videos' && <CardList videos={videos} />}
+            {showContent === 'Podcasts' && <CardList podcasts={podcasts} />}
+
```
jestapinski, line 44: Nice work getting this logic more clean!

src/components/dev-hub/tab.js
```diff
@@ -0,0 +1,79 @@
...
+    border-bottom: 1px solid ${colorMap.greyDarkOne};
+`;
+
+const activeStyles = css`
+    color: ${colorMap.devWhite};
+    border-bottom-color: ${colorMap.darkGreen};
+`;
+
+const defaultStyles = css`
+    &:hover {
```
jestapinski, line 18: for the `:hover`, we should be able to add it into a `styled(P)`, does that work?

tahiya-mongo: Hovering does work. However, because we are separating the text from the button when we do that, the text color doesn't seem to remain white (indicating it's active) once clicked.

src/pages/media.js
tahiya-mongo: Thanks for pointing that out!

src/components/dev-hub/tab.js
```diff
@@ -0,0 +1,78 @@
...
+    font-family: 'Fira Mono';
+
+    ${({ isActive }) => (isActive ? activeStyles : defaultStyles)}
+`;
+
+export default ({ handleClick, leftTabs, rightTabs, activeItem }) => {
+    return (
+        <Tab>
+            <div>
+                {leftTabs.map(item => {
```
jestapinski, line 48: Since this map is also the same as the right tab map, can we extract this out into some `mapTabTextToButton` helper?

src/pages/media.js
```diff
@@ -13,10 +14,34 @@ export default () => {
...
+    const leftTabs = ['All'];
+    const rightTabs = ['Articles', 'Videos', 'Podcasts'];
+
     return (
         <Layout>
+            <Tab
+                handleClick={display}
+                leftTabs={leftTabs}
+                rightTabs={rightTabs}
```
jestapinski, line 31: Nice!

src/components/dev-hub/tab.js
jestapinski: Hmm I tried adding a `P` and that seemed to do it? Here is what I had, does this work for you?

```
                        <TabButton
                            key={item}
                            isActive={isActive}
                            onClick={() => handleClick(item)}
                        >
                            <P collapse>{item}</P>
                        </TabButton>
```

jestapinski: Actually should this `defaultStyle` hover be applied to the button all the time? (even if active?) That can reduce complexity here

src/components/dev-hub/tab.js
```diff
@@ -0,0 +1,78 @@
...
+`;
+
+const TabButton = styled('button')`
+    border: none;
+    outline: none;
+    cursor: pointer;
+    transition: 0.3s;
+    background-color: inherit;
+    width: 136px;
+    padding-top: ${size.micro};
```
jestapinski, line 31: Nitpicky, but can we reduce this to [one-line padding css](https://www.w3schools.com/css/css_padding.asp): `padding: ${size.micro} ${size.medium} ${size.mediumLarge} ${size.mediumLarge}`

src/components/dev-hub/tab.js
jestapinski: Sorry to clarify can we take this `hover` stuff and put it in `TabButton` directly since it is applied at all times? Then we can remove this `defaultStyles` var and remove this logic also from `activeStyles`

src/components/dev-hub/tab.js
```diff
@@ -0,0 +1,68 @@
...
+    transition: 0.3s;
+    background-color: inherit;
+    width: 136px;
+    padding: ${size.micro} ${size.medium} ${size.mediumLarge}
+        ${size.mediumLarge};
+    border-bottom: 3px solid transparent;
+
+    font-size: ${fontSize.default};
+    color: ${colorMap.greyDarkOne};
+    font-family: 'Fira Mono';
```
jestapinski, line 40: Since `P` works now are these lines all still needed? Color might be but I suspect font-family and size are not

src/pages/media.js
```diff
@@ -13,10 +14,34 @@ export default () => {
         isLoading: isLoadingPodcasts,
     } = usePodcasts();
 
+    const [activeItem, setActiveItem] = useState('All');
+
+    const display = content => {
```
jestapinski, line 19: Actually we can get away without `display` since it is just `setActiveItem`. Let's pass `setActiveItem` in as a prop for `handleClick` directly.

src/components/dev-hub/tab.js
tahiya-mongo: It looks like we do still need them. `P` doesn't specify a font-family and the size is smaller than the one specified in the design.

src/components/dev-hub/tab.js
```diff
@@ -28,10 +31,8 @@ const TabButton = styled('button')`
     transition: 0.3s;
     background-color: inherit;
     width: 136px;
-    padding-top: ${size.micro};
-    padding-bottom: ${size.medium};
-    padding-left: ${size.mediumLarge};
-    padding-right: ${size.mediumLarge};
+    padding: ${size.micro} ${size.medium} ${size.mediumLarge}
```
jestapinski, line 34: Whoops just noticed this! With the padding shorthand it is actually `top right bottom left` so the correct use here would be `padding: ${size.micro} ${size.mediumLarge} ${size.medium} ${size.mediumLarge}`.

Even better there is a shorter shorthand which is `padding: <top> <left and right> <bottom>`, so `padding: ${size.micro} ${size.mediumLarge} ${size.medium}` is also equivalent!

src/components/dev-hub/tab.js
```diff
@@ -41,38 +42,27 @@ const TabButton = styled('button')`
...
+            <TabButton
+                key={item}
+                isActive={isActive}
+                onClick={() => handleClick(item)}
+            >
+                <P collapse>{item}</P>
+            </TabButton>
+        );
+    });
+};
```
jestapinski, line 58: Super nit but if we are returning right away from a function (first thing we do is return), we can omit the `return` keyword and curly braces like so:

```
const mapTabTextToButton = (textList, activeItem, handleClick) => (
   textList.map(item => {
        const isActive = item === activeItem;
        return (
            <TabButton
                key={item}
                isActive={isActive}
                onClick={() => handleClick(item)}
            >
                <P collapse>{item}</P>
            </TabButton>
        );
    });
);
```

Just makes things a little cleaner!

src/components/dev-hub/tab.js
jestapinski: Ah okay! Thanks for digging into this

src/pages/media.js
```diff
@@ -1,9 +1,10 @@
-import React from 'react';
+import React, { useState } from 'react';
 import Layout from '../components/dev-hub/layout';
+import Tab from '../components/dev-hub/tab';
 import useAllVideos from '../hooks/use-all-videos';
-import { P } from '../components/dev-hub/text';
-import CardList from '../components/dev-hub/card-list';
 import usePodcasts from '../hooks/use-podcasts';
+import { P } from '../components/dev-hub/text';
+import CardList from '../components/dev-hub//card-list';
```
jestapinski, line 7:
```suggestion
import CardList from '../components/dev-hub/card-list';
```

Looks like an extra / here

src/components/dev-hub/tab.js
```diff
@@ -0,0 +1,57 @@
...
+    border-bottom: 3px solid transparent;
+    color: ${colorMap.greyDarkOne};
+    cursor: pointer;
+    font-size: ${fontSize.default};
+    font-family: 'Fira Mono';
+    &:hover {
+        color: ${colorMap.devWhite};
+    }
+    outline: none;
+    padding: ${size.micro} ${size.mediumLarge} ${size.medium};
```
jestapinski, line 32: Just checked one last time and the bottom padding is only 16px! Would you mind updating that using `theme.js`?

tahiya-mongo: Not a problem!

PR #356, opened 2020-05-04
Gb/rss tweaks

greggb:
Changed title to just the site title, and added a valid image type


.gitignore
```diff
@@ -2,9 +2,6 @@ node_modules/
 public/
 .cache/
 .DS_Store
-static/*
```
jestapinski, line 5: We don't want to accidentally commit all of the statics accidentally after building, can we special case this one file and keep the existing gitignore rules?

greggb: Going to message you about this - have a question about why this is necessary

greggb:
waiting for twitch api to be back up before re-running

PR #357, opened 2020-05-04
Add Video Props to Card

jestapinski:
To prep for upcoming work on cards with videos, this PR adds a `VideoCard` component which wraps the card to open a video modal, etc on click.

One key note is we need to absolutely position items on the `image` of this card, so I added a prop to render children in a specific place in the `Card` to do so. I'll comment in the code.

Here is how it looks with more card details (the play button is correctly centered):
Screen Shot 2020-05-04 at 1 45 36 PM

Screen Shot 2020-05-04 at 1 45 44 PM


src/pages/index.js
```diff
@@ -94,37 +93,6 @@ const DescriptiveText = styled(P)`
...
-const ThumbnailCard = styled(Card)`
-    position: relative;
-`;
-
-const ThumbnailButton = styled(Button)`
-    left: 0;
-    right: 0;
-    top: 0;
-    /* Account for bottom margin and title of card */
-    bottom: 52px;
```
jestapinski, line 106: This value for centering the play button was hardcoded for the use case on the index page, so we needed a more general solution for use in other places

src/components/dev-hub/card.js
```diff
@@ -127,8 +110,10 @@ const Card = ({
         >
             <div>
                 {!collapseImage && (
-                    <ImageWrapper>
+                    <ImageWrapper placeChildrenOnImage={placeChildrenOnImage}>
                         {image && <Image src={image} alt="" />}
+                        {/* Expect children to be absolutely positioned on the image */}
+                        {placeChildrenOnImage && children}
```
jestapinski, line 116: Here is where we toggle where children are rendered. Another note is that we don't really use `children` anywhere as-is so this bool isn't 100% necessary, however since it isn't the typical card-child behavior it might be worth keeping.

src/components/dev-hub/card.js
```diff
@@ -69,26 +71,6 @@ const CardTitle = styled(H5)`
     text-align: left;
 `;
 
-/**
```
jestapinski, line 72: deprecated

src/components/dev-hub/card.js
```diff
@@ -27,6 +27,8 @@ const ImageWrapper = styled('div')`
     overflow: hidden;
     padding: 0;
     width: 100%;
+    ${({ placeChildrenOnImage }) =>
```
tahiya-mongo, line 30: Nice! Didn't know you could do that.

src/components/dev-hub/card.js
greggb: This is being used to insert the video modal component inside the imagewrapper?

jestapinski: This is where the video button is being placed inside the image wrapper. Another idea would just be to move the video button here and toggle it with a prop actually, how does that sound?

(That way `children` still behaves consistently)

jestapinski:
I changed my approach here to just add some new props to the `Card` component to allow for a `VideoModal` to be supported. By adding `video` and `videoModalThumbnail` we can allow any `Card` to play videos in the modal.

Still looks fine with other card props, including the new `badge` on the image.

jestapinski:
Thanks all!

PR #359, opened 2020-05-05
Remove Prefix Path from Cypress Actions

jestapinski:
This PR creates new package scripts: `buildTest` and `serveTest` which remove references to `prefix-paths` to ensure cypress runs are consistent between branches in CI.


jestapinski:
Thanks!

PR #362, opened 2020-05-06
Update Deps

jestapinski:
I noticed in `package.json` none of our deps were pinned, so running `npm i` could result in inconsistent lockfiles.

This PR ~pins deps and also~ updates most of them with minor version bumps.

~Pinning deps should help with inconsistencies in Github Actions failing occasionally on `npm ci`.~ 

Also references to `node-pre-gyp` in the lockfile are gone, which had some issues with `npm ci` in Github Actions.


greggb:
I think this is fine, but wanted to ask about a couple things.
- could unpinning and using [`npm ci`](https://docs.npmjs.com/cli/ci.html) accomplish the same thing?
- Alternatively, should we consider using yarn here like we do in UP with a `frozen-lockfile` - the lockfile churn was a big reason I moved us to yarn there.

jestapinski:
> * could unpinning and using [`npm ci`](https://docs.npmjs.com/cli/ci.html) accomplish the same thing?

Yes, the reason this happened in the first place was the bad workperpool job was using `npm install` instead of `npm ci` iirc. Ideally we should always use `npm ci` except for dep updates.

> * Alternatively, should we consider using yarn here like we do in UP with a `frozen-lockfile` - the lockfile churn was a big reason I moved us to yarn there.

I agree, would this require the DOCSP team to update the workerpool jobs as well?

greggb:
>> * could unpinning and using [`npm ci`](https://docs.npmjs.com/cli/ci.html) accomplish the same thing? ...

@jestapinski we’d only need to do one thing. If we can guarantee ci usage we can avoid having to update to yarn in the pool jobs (and getting yarn installed there)

jestapinski:
> > > * could unpinning and using [`npm ci`](https://docs.npmjs.com/cli/ci.html) accomplish the same thing? ...
> 
> @jestapinski we’d only need to do one thing. If we can guarantee ci usage we can avoid having to update to yarn in the pool jobs (and getting yarn installed there)

That makes sense, I'll slack you some other questions on it

jestapinski:
Thanks for the help!

PR #363, opened 2020-05-06
WIP gatsby-source-mongodb-stitch

jestapinski:
This PR creates a very basic Gatsby source plugin for MongoDB Stitch use at build-time. Right now, the plugin is local, but the idea would be to separate it and register it with `npm`, add it to the [Gatsby plugin directory](https://www.gatsbyjs.org/plugins/), etc. It would also likely need some more functionality, but I need to learn more about Stitch to figure out what might make sense to add.

Specifically with our site, there are some pros and cons that might be worth calling out:

Pros:
- Moves all of our `sourceNodes` logic out and into the plugin
- Forces us to be more Gatsby by working with nodes for our foundational pieces of data
- "Eat your own dog food" 

Cons:
- Still need to call Stitch in other places during build-time, and we cannot move these calls into graphql so we cannot use them in the plugin
- Overhead

Open Questions:
- If we were to move forward with this plugin, we will likely need to add some more functionality. Not sure what more we would want to do?


jestapinski:
twitch is down again so CI is failing for that

jestapinski:
Closing this since the plugin is going to be hosted separately soon and we will just integrate here when it is up

PR #364, opened 2020-05-07
Remove live API key, omit twitch test

jestapinski:
This PR does two things:
- Removes the live API key to fix API calls to live.mongodb.com for events
- Omits the twitch-related cypress tests while we figure out the API issues there


jestapinski:
Thanks for taking a look!

PR #367, opened 2020-05-07
Create an iframe twitch fallback component

jestapinski:
This PR creates a component used if Twitch API calls fail/give us `null` video data. A question on this iframe:
- Using the channel id always checks for streams, there is no way to just get the last video without an ID. We could hardcode a video ID since more often than not the team won't be streaming, does this make sense here?

Desktop:
Screen Shot 2020-05-07 at 5 11 01 PM

Mobile:
Screen Shot 2020-05-07 at 5 11 18 PM

### DIFF
--- a/src/components/dev-hub/twitch-fallback-card.js
+++ b/src/components/dev-hub/twitch-fallback-card.js
@@ -1,0 +1,33 @@
+import React from 'react';
+import styled from '@emotion/styled';
+import Card from './card';
+import { screenSize } from './theme';
+
+const StyledIframe = styled('iframe')`
+    height: 100%;
+    width: 100%;
+    border: none;
+`;
+
+const TwitchCard = styled(Card)`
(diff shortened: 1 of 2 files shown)


src/components/dev-hub/twitch-fallback-card.js
```diff
@@ -0,0 +1,33 @@
...
+    height: 100%;
+    width: 100%;
+    border: none;
+`;
+
+const TwitchCard = styled(Card)`
+    height: 100%;
+    width: 100%;
+    @media ${screenSize.upToLarge} {
+        height: 300px;
```
jestapinski, line 16: Was getting some percentage responsiveness issues since this isnt an `img` but didn't think it was necessary to dig into the media block and card for this one case

jestapinski:
Sounds good, thanks!

PR #369, opened 2020-05-08
Add lazy loading to article, card images

jestapinski:
This PR adds the `loading="lazy"` attribute to:
- Images within articles
- Images on cards

I also checked out the network tab and verified it didn't load certain images until they were close to the viewport, neat!

Some snapshots needed to be updated due to the new attribute


jestapinski:
Thanks!

PR #371, opened 2020-05-11
Move YT API Key into env files

jestapinski:
**This would require DOCSP to add the YT_API_KEY in the .env files in jobs**

This PR pulls the YouTube API Key out of the fetch util and reads from the `.env.*` files. Not sure which other API keys we would want to move out? This was the only one I saw (Twitch has ClientID but most of that code will likely move out of our repo soon anyway).


jestapinski:
Going to do this via Stitch

PR #372, opened 2020-05-11
Request YouTube Playlists from a Second Stitch App

jestapinski:
This PR moves our logic for requesting YouTube playlists to a function hosted in a new MongoDB Stitch App. You all should have gotten an invite to a stitch app called `devhub-frontend` and a project called `devhub-authentication`, which will be using also for Twitch Authentication (we can rename later if desired).

We were getting flagged for having a public key in the repo, so moving this into Stitch both gives us the chance to play with Stitch (for Twitch later) and clean up these warnings.

How this works is instead of a `fetch` HTTP request we `await` a request to Stitch. The code for handling a youtube playlist is on Stitch called `fetchYoutubeData` and is almost identical to what we had before.

We were using the snooty Stitch app for submitting projects but didn't want to touch that functionality if possible.


src/utils/stitch.js
```diff
@@ -35,3 +38,15 @@ export const callStitchFunction = async (fnName, metadata, ...fnArgs) => {
         console.error(error);
     }
 };
+
+export const callStitchAuthenticationFunction = async (
```
jestapinski, line 42: This was messy, there is another function right above this which is a wrapper for Stitch but needed some more work to get `metadata` and supporting different `appIds` sorted out, hence still WIP but would like to get feedback

greggb: I'm wondering if we should separate the clients so we have one for content and one for APIs.

We could keep these separated so the logic on what each is doing is more clear.

jestapinski: If by `client` you mean a different app on Stitch, then yes we have different apps (the `snooty` one handles content, the `devhub-authentication` one is new and will be for APIs).

Is this what you were referring to?

jestapinski: Here I was just trying to re-use the Stitch-related code to support different `appId`s

greggb: I meant more like `const contentClient` and `const apiClient` so we don't have to think about the logic between the two apps

jestapinski: Oh I see, keeping them more intentionally separate, that makes some sense.

I'd like to try having a `stitch.js` with some common Stitch commands and then two modules which extend them, `snootyStitch.js` and `devhubApiStitch.js` to help solidify this distinction.

jestapinski:
🎉

PR #373, opened 2020-05-12
Switch 'target' to 'fileid' for related articles

jestapinski:
This solves half of the problem for [DEVHUB-124](https://jira.mongodb.org/browse/DEVHUB-124).

Looks like the data of returned articles docs either changed or was different from what was expected. Locally I was able to recreate the issue. On the newest parser version (0.4.6) this is what data we get for a related `doc`:

Screen Shot 2020-05-12 at 11 22 14 AM

Unfortunately, the image is blank and so is target. Since fileid and target are identical I switched to using `fileid` which should solve the related article not showing up. As for the image, I can dig into this some more as well.


jestapinski:
Closing since there is an issue at buildtime instead

PR #374, opened 2020-05-12
Fix related pages by using refuri and fileid

jestapinski:
This PR solves [DEVHUB-124](https://jira.mongodb.org/browse/DEVHUB-124).

The issue was the data we were getting back for related articles was different from what we previously had. Before we just got `target` for all types of article links (internal and external). Now they are sending multiple types of related articles back with different fields for each based on whether they use the `doc` directive or try as a link (what Lauren was doing). This PR will handle both cases for robustness:

A "reference" type
Screen Shot 2020-05-12 at 12 15 08 PM

A "ref_role" type
Screen Shot 2020-05-12 at 12 15 17 PM

Neither of these had `target` so building related articles would fail. Also, these values have leading slashes, which older values did not.

This PR optionally checks for any possible field that could have the slug/url: `target` (not sure if this is even in use anymore), `refuri`, and `fileid` to account for different types. It also removes the leading slash to solve issues getting the title and image and uses the `image` should it exist for related articles of any type (`reference` or `ref-doc`) should the content writers prefer one syntax over the other. 

The tests also had to be updated for this change. I also plan to add a cypress test for these images in a bit more detail.

This PR:
Screen Shot 2020-05-12 at 12 20 52 PM

### DIFF
--- a/src/utils/setup/get-related-pages-with-images.js
+++ b/src/utils/setup/get-related-pages-with-images.js
@@ -5,7 +5,7 @@ const getRelatedPagesWithImages = (pageNodes, RESOLVED_REF_DOC_MAPPING) => {
     const relatedPageInfo = related.map(r => {
         // Handle `reference` and `ref_role` types
         const target = r.target || r.refuri || r.fileid;
-        const slug = target && target.slice(1, target.length);
+        const slug = target && target.slice(1);
         return {
             ...r,
             target,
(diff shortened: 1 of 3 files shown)


src/utils/setup/get-related-pages-with-images.js
```diff
@@ -2,14 +2,20 @@ const dlv = require('dlv');
...
-            RESOLVED_REF_DOC_MAPPING,
-            [r.target, 'query_fields', 'atf-image'],
-            null
-        ),
-        ...r,
-    }));
+    const relatedPageInfo = related.map(r => {
+        // Handle `reference` and `ref_role` types
+        const target = r.target || r.refuri || r.fileid;
+        const slug = target && target.slice(1, target.length);
```
greggb, line 8: nit
```suggestion
        const slug = target && target.slice(1);
```

jestapinski:
Thanks! Found an issue with `withPrefix` so added that in to handle image sources for atf images. Will verify again in staging.

PR #378, opened 2020-05-13
Use gatsby-source-mongodb-stitch

jestapinski:
Recently a [new plugin to support Stitch functions](https://www.npmjs.com/package/gatsby-source-mongodb-stitch) at Gatsby build-time came out, so this PR adds support for this plugin to the devhub. The plugin is nice because it cuts some logic out of `gatsby-node` and forces us to continue to use the concept of nodes.

There should be no change in functionality.


greggb:
> Recently a [new plugin to support Stitch functions](https://www.npmjs.com/package/gatsby-source-mongodb-stitch) at Gatsby build-time came out

🤔

PR #379, opened 2020-05-13
[DO NOT MERGE] Add Tab Component to the Learn Page

tahiya-mongo:
This PR adds the tab component to the `Learn` page. Users will eventually be able to toggle between all the different MongoDB related content on the `Learn` page. 

The filter bar was left alone (it only works for articles) for now until we have more information.


src/pages/learn.js
```diff
@@ -65,12 +69,19 @@ const HeaderContent = styled('div')`
     margin-right: auto;
 `;
 
-const Article = styled('article')`
+const CardContainer = styled('div')`
```
jestapinski, line 72: Can we undo this? using `article` is a good practice for a main piece of content

src/pages/learn.js
```diff
@@ -65,12 +69,19 @@ const HeaderContent = styled('div')`
...
     padding: ${size.medium};
 `;
 
 const StyledFilterBar = styled(FilterBar)`
-    padding-bottom: ${size.medium};
+    margin: 0 auto;
+    padding-bottom: ${size.large};
+    width: 95%;
```
jestapinski, line 79: using % widths aside from say 100 have some unintended effects when it comes to being responsive and with the filter bar at 94 I'm sure we can think about this a bit more. It would be safer for us to use some padding instead, would that get the job done?

src/pages/learn.js
```diff
@@ -211,14 +240,37 @@ export default ({
...
-            <Article>
+
+            <TabBar
+                handleClick={setActiveItem}
+                leftTabs={leftTabs}
+                rightTabs={rightTabs}
+                activeItem={activeItem}
+            />
+
+            {(isLoadingVideos || isLoadingPodcasts) && <P>Loading...</P>}
```
jestapinski, line 251: We might actually want to leave this out for now since it could look odd having articles loaded (since they are given at build time) then it still saying "loading...". I'll ask Dan about this but let's omit it for now if that is alright.

src/pages/learn.js
```diff
@@ -211,14 +240,37 @@ export default ({
...
+                        videos={videos}
+                        podcasts={podcasts}
+                    />
+                )}
+                {activeItem === 'Articles' && <CardList articles={articles} />}
+                {activeItem === 'Videos' && <CardList videos={videos} />}
+                {activeItem === 'Podcasts' && <CardList podcasts={podcasts} />}
+            </CardContainer>
+
+            {(errorVideos || errorPodcasts) && (
```
jestapinski, line 271: Similar thing here, even if these fail we still have articles. Another Dan question for later but let's omit for now since *some* content exists.

src/pages/learn.js
```diff
@@ -211,14 +240,37 @@ export default ({
...
+                {activeItem === 'All' && (
+                    <CardList
+                        articles={articles}
+                        videos={videos}
+                        podcasts={podcasts}
+                    />
+                )}
+                {activeItem === 'Articles' && <CardList articles={articles} />}
+                {activeItem === 'Videos' && <CardList videos={videos} />}
+                {activeItem === 'Podcasts' && <CardList podcasts={podcasts} />}
```
jestapinski, line 268: I also need to get clarification on what the filters should do in these cases

src/components/dev-hub/filter-bar.js
```diff
@@ -27,7 +27,7 @@ const zipFilterObjects = (filterObject, findCount = x => x.count) => {
 const ResponsiveFlexContainer = styled('div')`
     align-items: center;
     display: flex;
-    justify-content: space-between;
+    margin-left: auto;
```
jestapinski, line 30: We can actually omit this and use flexbox to do the same thing (since we are already using flexbox with the parent).

In the parent css we can set `justify-content: flex-end` which will magically push the filters to the right! That way we can keep working in flexbox here!

src/pages/learn.js
```diff
@@ -69,19 +69,17 @@ const HeaderContent = styled('div')`
...
 `;
 
-const CardContainer = styled('div')`
+const Article = styled('article')`
     padding: ${size.medium};
 `;
 
 const StyledFilterBar = styled(FilterBar)`
-    margin: 0 auto;
+    margin: 0 34px;
```
jestapinski, line 77: Are these px coming from the design? If so can we just use `large` from theme.js? How does that look?

tahiya-mongo: They're not. I was just trying out different px to align the cards, filter, and tab component based on the design. I tried using `large` from theme.js first, but things weren't perfectly aligned. If that's not a hard requirement, I can go ahead and change it to `large`.

jestapinski: Yeah let's try to keep to theme.js, I spoke to Dan a bit about which sizes we have for the future so let's use `large` here (or something from `theme`) if possible!

tahiya-mongo:
Sounds good! Thanks for the feedback so far

src/components/dev-hub/card-list.js
```diff
@@ -62,7 +61,7 @@ const renderArticle = article => (
 
 const renderVideo = video => (
     <VideoCard
-        key={video.title}
+        key={video.mediaType + video.title}
```
jestapinski, line 64: I had to update these because titles were conflicting. For example, the team is uploading podcasts to youtube with the same title. This caused some crazy behavior. Mashing the type of media with title is safer.

jestapinski:
🚢

PR #380, opened 2020-05-13
Stitch with Twitch (Videos)

jestapinski:
This PR supplements work done in the new Stitch app to comply with the new authentication requirements for the Twitch API and request for videos (not streams, although that would look very similar and finish this task I decided to just focus on videos here).

The code change here is pretty small. We pull out work related to `fetch`ing Twitch videos since this is now being done via Stitch and just request a new Stitch function `fetchMDBTwitchVideos` for the content.

In Stitch I aded a few things. In addition to the `fetchMDBTwitchVideos` function for content, I added a function called `fetchTwitchAppToken` this uses the client ID and secret (stored in `Values & Secrets`) to get an [App Access Token](https://dev.twitch.tv/docs/authentication/#types-of-tokens) to Twitch. This token is active for ~60 days so we need not continuously generate tokens. That being said we can call Twitch's validate endpoint to make sure the token is valid, but I'm not sure how often we should be doing so.

If you check out `/media`, you should see Twitch videos again.

Open Questions:
- I can't think of a way to automatically update the App Access token when it is expired or close to expire. Ideally I would like a way to programmatically roll our secrets, but not sure if this is supported. Since it is about a two month window before we would need to generate a new token, perhaps setting up a monthly trigger to tell us to generate a token wouldn't be terrible, but it seems less than ideal. We can also setup a Stitch Trigger to check `/validate` every so often and alert us if it goes down, but I would hope for a more preventative measure. Thoughts?
- How often should we check Twitch's [validate endpoint](https://dev.twitch.tv/docs/authentication/#validating-requests)? It says "before each mutation request or accessing sensitive data" but I don't believe the data we are accessing is that sensitive.


jestapinski:
thanks!

PR #382, opened 2020-05-14
Use Stitch for Twitch Streams

jestapinski:
This PR is the last piece of work for moving existing APIs to Stitch, and also finishes migrating our Twitch use to the New Twitch API.

This PR incorporates a new Stitch function called `fetchMDBTwitchStream`, which is very similar to work done for Twitch Videos into the codebase by adding a new wrapper to call Stitch and incorporating it in the hook used by the homepage to check streams before checking videos.

Now, on the homepage we should see the old card back again (although there is no stream right now, a video does show) and no 401 errors in the console. The title was also missing from this card so I added it back in.

### DIFF
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -152,6 +152,7 @@ export default ({ pageContext: { featuredArticles } }) => {
                                         twitchVideo.thumbnailUrl
                                     )}
                                     maxWidth={MEDIA_WIDTH}
+                                    title={twitchVideo.title}
                                     videoModalThumbnail={getTwitchThumbnail(
                                         twitchVideo.thumbnailUrl,
                                         1200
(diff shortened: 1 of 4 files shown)


src/pages/index.js
```diff
@@ -152,6 +152,7 @@ export default ({ pageContext: { featuredArticles } }) => {
                                         twitchVideo.thumbnailUrl
                                     )}
                                     maxWidth={MEDIA_WIDTH}
+                                    title={twitchVideo.title}
```
greggb, line 155: When did this get removed, and was it on purpose? I know we had some design changes before launch that might have included removing the title

jestapinski: This was removed by accident about a week ago when I added video props to the `Card` component. I removed the `Thumbnail` component from the index page in lieu of just using `Card` here instead at that point. [This is the PR](https://github.com/mongodb/devhub/pull/357/files) where that change happened.

jestapinski:
Thanks!

PR #384, opened 2020-05-14
Storybook entry for the badge component and card-list changes

tahiya-mongo:
This PR documents the `Badge` component on storybook and makes the thumbnail images for twitch videos square to match the other cards (Youtube thumbnails are still in progress). It also renames the `CardBadge` component to `Badge` since its use is not restricted to just `Card` components.

### DIFF
--- a/src/components/dev-hub/badge.js
+++ b/src/components/dev-hub/badge.js
@@ -3,7 +3,7 @@ import { colorMap, size } from './theme';
 import { P5 } from './text';
 import styled from '@emotion/styled';
 
-const CardBadge = styled('div')`
+const Badge = styled('div')`
     bottom: 0;
     background-color: ${colorMap.greyDarkThree};
     font-family: akzidenz;
@@ -35,9 +35,9 @@ const badgeContent = contentType =>
         : contentType;
 
 export default ({ contentType, color = '' }) => (
-    <CardBadge color={color || determineColor(contentType)}>
+    <Badge color={color || determineColor(contentType)}>
         <P5 bold collapse>
             {badgeContent(contentType)}
         </P5>
-    </CardBadge>
+    </Badge>
 );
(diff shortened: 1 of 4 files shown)


src/components/dev-hub/badge.js
```diff
@@ -35,9 +35,9 @@ const badgeContent = contentType =>
         : contentType;
 
 export default ({ contentType, color = '' }) => (
-    <CardBadge color={color || determineColor(contentType)}>
+    <Badge color={color || determineColor(contentType)}>
```
jestapinski, line 38: I think this name change makes sense!

src/components/dev-hub/stories/badge.stories.mdx
```diff
@@ -0,0 +1,53 @@
...
+import Badge from '../badge';
+import { colorMap } from '../theme';
+
+<Meta title="Badge" />
+
+# Badge
+
+Badges can be used to label contents. It accepts two parameters: `contentType` and `color`.
+* The `contentType` variable defines the text that goes inside the badge.
+* The optional `color` variable defines the color of the borders around the badge.
```
jestapinski, line 11: Can we add a note on how it is `position: absolute` by default? Definitely good to know in case someone wants to use this otherwise.

src/components/dev-hub/stories/badge.stories.mdx
```diff
@@ -0,0 +1,53 @@
...
+
+<Preview>
+    <Story name="custom-badge">
+        <div style={{position:'relative', height: '50px'}}>
+            <Badge contentType={'badge'} color={colorMap.teal}/>
```
jestapinski, line 16: I noticed the font is off a bit but this is likely something with the storybook configuration. Don't sweat this I will look into it more separately.

PR #385, opened 2020-05-14
List tab buttons in a single column on mobile

tahiya-mongo:
This PR makes minor css changes to list the buttons on the `Tab` component in a single column (as opposed two) on mobiles.

### DIFF
--- a/src/components/dev-hub/tab.js
+++ b/src/components/dev-hub/tab.js
@@ -10,7 +10,7 @@ const Tab = styled('div')`
     border-bottom: 1px solid ${colorMap.greyDarkOne};
     display: flex;
     justify-content: space-between;
-    @media ${screenSize.upToLarge} {
+    @media ${screenSize.upToMedium} {
         display: block;
     }
 `;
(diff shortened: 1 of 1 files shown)


src/components/dev-hub/tab.js
```diff
@@ -10,6 +10,9 @@ const Tab = styled('div')`
     border-bottom: 1px solid ${colorMap.greyDarkOne};
     display: flex;
     justify-content: space-between;
+    @media ${screenSize.upToLarge} {
```
jestapinski, line 13: This works! Didn't even think of just removing flexbox :)

Can break at `medium` here instead of `large` since that is where most of our mobile styles kick in (nav, single card per row, etc) and below?

src/components/dev-hub/tab.js
```diff
@@ -33,6 +36,10 @@ const TabButton = styled('button')`
     transition: 0.3s;
     width: ${TAB_WIDTH};
     ${({ isActive }) => isActive && activeStyles}
+    @media ${screenSize.upToLarge} {
+        display: block;
+        margin:0 auto;
```
jestapinski, line 41:
```suggestion
        margin: 0 auto;
```
super nit

PR #386, opened 2020-05-15
Create a Slider Component

jestapinski:
This PR creates a Slider component using `<input type="range" />`, which we will use when playing podcasts. It supports props for state management by parents, and works on chrome, ff, and safari!

I also made a quick story for it.

Screen Shot 2020-05-15 at 3 31 02 PM


jestapinski:
> I think getting the player playing the podcasts on the page would be a good next step here before building out too much more UI. Custom players can be somewhat tricky and we might find we need a different approach or just use the default UI

Yeah working on this now, fortunately it looks good so far, should have something to show soon

jestapinski:
Thanks!

PR #389, opened 2020-05-18
Basic Audio Component

jestapinski:
This PR adds an audio player for podcasts using `ReactPlayer` underneath a custom UI. It also works on all browsers. It can be seen by going to `/media` and clicking a podcast.

Here are some basic screenshots:
Screen Shot 2020-05-18 at 4 15 01 PM
Screen Shot 2020-05-18 at 4 15 10 PM
Screen Shot 2020-05-18 at 4 16 10 PM
Screen Shot 2020-05-18 at 4 16 16 PM


jestapinski:
This PR is ready for review. The designs got tricky because of what we want to show then, and having a different scheme for mobile. A css grid approach made the most sense to me to allow for us to re-organize quickly but there may be a cleaner way to do the same.

src/components/dev-hub/slider.js
```diff
@@ -56,7 +56,11 @@ const StyledInput = styled('input')`
 
 const SliderLabelText = styled(P3)`
     font-family: 'Fira Mono';
-    margin: 0 ${size.default};
+    margin: 0 ${size.default} 0 0;
```
jestapinski, line 59: This prevents unnecessary extra margins on the labels to the slider

src/components/dev-hub/audio.js
```diff
@@ -0,0 +1,259 @@
+import React, { useCallback, useEffect, useState } from 'react';
```
jestapinski, line 1: This component file is very large, but it didn't seem to make sense to break this out any further since everything is pretty unique to this component right now.

jestapinski:
> This is looking good. I'm having some trouble tabbing between the main actions. I can get to the slider, but not the close, collapse or play/pause button.
> 
> As you mentioned there's a pretty heavy delay when clicking on a slider location. Is there a root cause for that?

I'll double check tab order on FF but my guess for Slider delay would be the audio buffering. Some of these episodes are ~30 minutes so it could just be slow to download. `ReactPlayer` does let us hook into how much is buffered so I can check this out as well.

jestapinski:
> This is looking really good! I noticed that if I visit any other tab while the player is playing it stops and immediately goes away. Is that the expected behavior?

Yes it is, we talked a bit about having the player "follow you" around but this is out of scope for the project at this point

src/components/dev-hub/audio.js
```diff
@@ -0,0 +1,256 @@
...
+    justify-content: space-between;
+    width: 100%;
+    a {
+        padding-left: 0;
+        padding-right: 0;
+        width: ${size.large};
+    }
+`;
+
+const AudioPlayButton = React.memo(({ isPlaying, toggleIsPlaying }) => (
```
jestapinski, line 143: Turns out the tab issues were that this components (and the options) kept being rerendered unnecessarily each second when hooks updated. I memoized these which prevented the extra rerenders and tabbing stays even after the slider updates.

jestapinski:
Thanks all!

PR #391, opened 2020-05-19
Add leafygreen checkbox

jestapinski:
For forms, we are also going to be using the leafygreen checkbox. This PR adds this package in and also creates a quick story for it. Should be simple enough to use with our forms.

Screen Shot 2020-05-19 at 9 56 41 AM


jestapinski:
Thanks!

PR #392, opened 2020-05-19
Section header component

tahiya-mongo:
This PR adds the section header component that will be used in the `Academia` page. 

Screen Shot 2020-05-19 at 10 32 07 AM

### DIFF
--- a/src/components/dev-hub/section-header.js
+++ b/src/components/dev-hub/section-header.js
@@ -1,0 +1,22 @@
+import React from 'react';
+import styled from '@emotion/styled';
+import { colorMap, fontSize, screenSize } from './theme';
+import { P } from './text';
+
+const SectionHeader = styled('div')`
+    border-bottom: 1px solid ${colorMap.greyDarkOne};
+    font-size: ${fontSize.small};
+    letter-spacing: 2px;
+    text-transform: uppercase;
+    @media ${screenSize.upToMedium} {
+        display: block;
+    }
+`;
+
+export default ({ text }) => (
+    <SectionHeader>
+        <div>
+            <P>{text}</P>
+        </div>
+    </SectionHeader>
+);
--- a/src/components/dev-hub/section-header.js
+++ b/src/components/dev-hub/section-header.js
@@ -7,16 +7,14 @@ const SectionHeader = styled('div')`
     border-bottom: 1px solid ${colorMap.greyDarkOne};
     font-size: ${fontSize.small};
     letter-spacing: 2px;
-    text-transform: uppercase;
     @media ${screenSize.upToMedium} {
         display: block;
     }
+    text-transform: uppercase;
 `;
 
 export default ({ text }) => (
     <SectionHeader>
-        <div>
-            <P>{text}</P>
-        </div>
+        <P>{text}</P>
     </SectionHeader>
 );
--- a/src/components/dev-hub/section-header.js
+++ b/src/components/dev-hub/section-header.js
@@ -1,19 +1,19 @@
 import React from 'react';
 import styled from '@emotion/styled';
 import { colorMap, lineHeight } from './theme';
-import { P } from './text';
+import { P4 } from './text';
 
 const SectionHeader = styled('div')`
     border-bottom: 1px solid ${colorMap.greyDarkOne};
     letter-spacing: 3px;
     text-transform: uppercase;
 `;
-const StyledP = styled(P)`
+const StyledP4 = styled(P4)`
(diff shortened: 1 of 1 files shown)


src/components/dev-hub/section-header.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import styled from '@emotion/styled';
+import { colorMap, fontSize, screenSize } from './theme';
+import { P } from './text';
+
+const SectionHeader = styled('div')`
```
jestapinski, line 6: Here are the specs for the text, I think we may want to use another `P` variant here to match since the default is 16px size and 24px line height. (`letter` is `letter-spacing`)

Screen Shot 2020-05-19 at 10 42 01 AM

src/components/dev-hub/section-header.js
```diff
@@ -0,0 +1,20 @@
...
+    border-bottom: 1px solid ${colorMap.greyDarkOne};
+    font-size: ${fontSize.small};
```
jestapinski, line 8: If we use another `P` I don't think we need this here either but we may need to `styled` that `P` to get the line height we want.

src/components/dev-hub/section-header.js
```diff
@@ -0,0 +1,20 @@
...
+    letter-spacing: 2px;
+    @media ${screenSize.upToMedium} {
+        display: block;
+    }
+    text-transform: uppercase;
+`;
+
+export default ({ text }) => (
+    <SectionHeader>
+        <P>{text}</P>
```
jestapinski, line 18:
```suggestion
        <P collapse>{text}</P>
```

Using the `collapse` prop will bring the border closer by removing the bottom margin on the `P` we decide to use.

src/components/dev-hub/section-header.js
```diff
@@ -0,0 +1,20 @@
...
+    border-bottom: 1px solid ${colorMap.greyDarkOne};
+    font-size: ${fontSize.small};
+    letter-spacing: 2px;
+    @media ${screenSize.upToMedium} {
+        display: block;
+    }
+    text-transform: uppercase;
+`;
+
+export default ({ text }) => (
```
jestapinski, line 16: Here is a cool trick! If we use a prop named `children`, then when we use this component we can do something like this:

```
<SectionHeader>For Academia</SectionHeader>
```

instead of

```
<SectionHeader text="For Academia" />
```

Personally I think the first one is more natural with html, but I'll leave it to you :)

src/components/dev-hub/section-header.js
```diff
@@ -0,0 +1,20 @@
...
+import { colorMap, fontSize, screenSize } from './theme';
+import { P } from './text';
+
+const SectionHeader = styled('div')`
+    border-bottom: 1px solid ${colorMap.greyDarkOne};
+    font-size: ${fontSize.small};
+    letter-spacing: 2px;
+    @media ${screenSize.upToMedium} {
+        display: block;
+    }
```
jestapinski, line 12: Do we need this?

src/components/dev-hub/section-header.js
tahiya-mongo: The first one definitely feels more natural. Thanks!

src/components/dev-hub/section-header.js
tahiya-mongo: You're right. We don't.

src/components/dev-hub/section-header.js
tahiya-mongo: The design seems to have bottom margin separating the text and the border

jestapinski: Ah yes there is `8px` between the text and the border. The updated `line-height` should give us this difference but if not good call let's keep that border. The gap seemed a bit high before but updating the text properties should help us.

tahiya-mongo: I left `collapse` out for now. Let me know what you think and I can change it if anything

jestapinski: Okay cool, yeah I think the default margin on `P` is `16px` so let's also add that to the `styled` and the font size should be `12px`. Once those are in I think this should be good to go

src/components/dev-hub/section-header.js
greggb: There's another consideration here - screen readers and search engines use the headers, `h1`, `h2`, etc to group the page and organize by importance. It would be good to keep these in importance order throughout the pages

tahiya-mongo:
Thanks!

PR #394, opened 2020-05-20
Emotion Theme Scaffolding

jestapinski:
This PR adds some scaffolding to support `emotion-theming` in anticipation of light mode coming later this quarter.

The approach I took was making a `darkTheme` object in `theme.js` and used that in `ThemeProvider`. Ideally when we wanted to make a `lightTheme` object it would be as easy as switching the theme passed to `ThemeProvider`. This also allows us to take an additive approach to bringing things into the theme, like we did for University.

I also switched the `layout` and `text` components to use the theme at this time.


jestapinski:
Thanks!

PR #396, opened 2020-05-20
1:1 Aspect Ratio For Any Card Image

jestapinski:
Before, we were always given square images for our cards, so we didn't have any issues with different aspect ratios causing different size cards. Unfortunately, YouTube thumbnails don't come in 1:1 aspect ratios, so we were getting this effect:

Screen Shot 2020-05-20 at 11 00 38 AM

This PR updates the Card Image logic to always fit the image into a 1:1 aspect ratio box so even if an image isn't a square, we will make it one. We take the initial image, center it, and fill the rest of the space so it matches other aspect ratio images. We get this done by using [padding-top against a 100% width](https://css-tricks.com/aspect-ratio-boxes/). We end up with this:

Screen Shot 2020-05-20 at 11 10 17 AM

### DIFF
--- a/src/components/dev-hub/card.js
+++ b/src/components/dev-hub/card.js
@@ -10,11 +10,15 @@ import VideoModal from './video-modal';
 import Badge from './badge';
 
 const Image = styled('img')`
+    bottom: 0;
     border-radius: ${size.small};
     display: block;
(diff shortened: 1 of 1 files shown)


jestapinski:
Thanks!

PR #398, opened 2020-05-20
Academia sign up form

tahiya-mongo:
This PR adds the UI for the sign up form that will allow instructors to join MongoDB Academia. It does not yet include the code to communicate with a backend upon submission. 

It also includes minor modification to both the `Select` and `SectionHeader` components to allow styling. 

Screen Shot 2020-05-20 at 4 13 34 PM

### DIFF
--- a/src/components/dev-hub/academia-sign-up-form.js
+++ b/src/components/dev-hub/academia-sign-up-form.js
@@ -14,10 +14,6 @@ const StyledButton = styled(Button)`
     margin: ${size.large} auto;
 `;
 
-// const StyledInput = styled(Input)`
-//     flex: 0 0 335px;
-// `;
-
 const StyledSelect = styled(Select)`
     color: ${colorMap.greyLightTwo};
 `;
--- a/src/components/dev-hub/academia-sign-up-form.js
+++ b/src/components/dev-hub/academia-sign-up-form.js
@@ -92,7 +92,7 @@ const Form = React.memo(({ setSuccess, success }) => {
     };
 
     return (
-        <Form onSubmit={handleSubmit}>
+        <form onSubmit={handleSubmit}>
             <ErrorMessage>
                 {success === false &&
                     'Your submission failed. Please try again.'}
@@ -189,7 +189,7 @@ const Form = React.memo(({ setSuccess, success }) => {
             >
                 Join MongoDB for Academia
             </StyledButton>
-        </Form>
+        </form>
     );
 });
 
--- a/src/components/dev-hub/academia-sign-up-form.js
+++ b/src/components/dev-hub/academia-sign-up-form.js
@@ -1,0 +1,213 @@
+import React, { useState } from 'react';
+import styled from '@emotion/styled';
+import { colorMap, size, screenSize } from './theme';
+import { P, ArticleH3, H3 } from './text';
+import Input from './input';
+import Button from './button';
+import FormSelect from './select';
+import Checkbox from '@leafygreen-ui/checkbox';
+import SectionHeader from './section-header';
+import SuccessIcon from './icons/success';
+
+const StyledButton = styled(Button)`
+    display: flex;
+    margin: ${size.large} auto;
+`;
+
+const StyledInput = styled(Input)`
+    height: 40px;
+    margin-bottom: ${size.large};
+    width: 335px;
+`;
+
+const StyledFormSelect = styled(FormSelect)`
+    color: ${colorMap.greyLightTwo};
+    height: 40px;
+    width: 335px;
+`;
+
+const SuccessContainer = styled('div')`
+    text-align: center;
+    svg {
+        margin-bottom: ${size.large};
+    }
+`;
+
+const ErrorMessage = styled(P)`
+    color: ${colorMap.salmon};
+`;
+
+const StyledSectionText = styled(ArticleH3)`
+    color: ${colorMap.greyLightTwo};
+    font-family: akzidenz;
+    font-weight: normal;
+`;
+
+const InstructorSection = styled('div')`
+    display: flex;
+    flex-wrap: wrap;
+    justify-content: space-between;
+    margin-bottom: ${size.default};
+    @media ${screenSize.upToMedium} {
+        display: block;
+    }
+`;
+
+const InstitutionSection = styled('div')`
+    display: flex;
+    justify-content: space-between;
+    margin-bottom: ${size.default};
+    @media ${screenSize.upToMedium} {
+        display: block;
+    }
+`;
+
+const StyledForm = styled('form')`
+    margin: 0 auto;
+    width: 720px;
+`;
+
+const Form = React.memo(({ setSuccess, success }) => {
+    const [firstName, setFirstName] = useState('');
+    const [lastName, setLastName] = useState('');
+    const [email, setEmail] = useState('');
+    const [institutionName, setInstitutionName] = useState('');
+    const [institutionType, setInstitutionType] = useState('');
+    const [agreeToEmail, setAgreeToEmail] = useState(false);
+    const [canSubmit, setCanSubmit] = useState(true);
+    const institutionTypes = [
+        ['Bootcamp', 'Bootcamp'],
+        ['Online Course', 'Online Course'],
+        ['High School', 'High School'],
+        ['University / College', 'University / College'],
+    ];
+
+    const handleSubmit = async e => {
+        e.preventDefault();
+        setCanSubmit(false);
+        const obj = {
+            first_name: firstName,
+            last_name: lastName,
+            email,
+            institution_name: institutionName,
+            institution_type: institutionType,
+            agree_to_email: agreeToEmail,
+        };
+
+        //TODO: communicate with the backend here
+    };
+
+    return (
+        <StyledForm onSubmit={handleSubmit}>
+            <ErrorMessage>
+                {success === false &&
+                    'Your submission failed. Please try again.'}
+            </ErrorMessage>
+
+            <SectionHeader>
+                <StyledSectionText>Instructor's Info</StyledSectionText>
+            </SectionHeader>
+
+            <InstructorSection>
+                <StyledInput
+                    value={firstName}
+                    maxLength="25"
+                    required
+                    pattern="^[A-Za-zÀ-ÿ ,.'-]+$"
+                    placeholder="First Name"
+                    onChange={e => setFirstName(e.target.value)}
+                    onInput={e => e.target.setCustomValidity('')}
+                    onInvalid={e =>
+                        e.target.setCustomValidity(
+                            'First names should only contain letters. e.g.John'
+                        )
+                    }
+                />
+                <StyledInput
+                    value={lastName}
+                    maxLength="25"
+                    required
+                    pattern="^[A-Za-zÀ-ÿ ,.'-]+$"
+                    placeholder="Last Name"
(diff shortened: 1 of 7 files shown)


src/components/dev-hub/academia-sign-up-form.js
```diff
@@ -0,0 +1,213 @@
...
+
+const StyledInput = styled(Input)`
+    height: 40px;
+    margin-bottom: ${size.large};
+    width: 335px;
+`;
+
+const StyledFormSelect = styled(FormSelect)`
+    color: ${colorMap.greyLightTwo};
+    height: 40px;
```
jestapinski, line 25: This actually breaks the dropdown, since we have to do some absolute positioning to get the effects we want. For the inputs can we just use `narrow` (even if it is a bit larger than the designs)? We can update the form elements at another time if this is what Dan really wants, but for now let's stick with `narrow` for all of them (aka remove `height: 40px` from styled form elements and just pass `narrow` as a prop).

jestapinski: This also gives some wonky effects when using the different form elements because they have some complicated css going on to get the gradient border and shifting up text effects.

src/components/dev-hub/academia-sign-up-form.js
```diff
@@ -0,0 +1,213 @@
...
+
+const Form = React.memo(({ setSuccess, success }) => {
+    const [firstName, setFirstName] = useState('');
+    const [lastName, setLastName] = useState('');
+    const [email, setEmail] = useState('');
+    const [institutionName, setInstitutionName] = useState('');
+    const [institutionType, setInstitutionType] = useState('');
+    const [agreeToEmail, setAgreeToEmail] = useState(false);
+    const [canSubmit, setCanSubmit] = useState(true);
+    const institutionTypes = [
```
jestapinski, line 78: Ah can we update the values (not the labels) here to be underscore_case and remove the slashes? I have a feeling that might be easier to work with on the backend.

i.e. "University / College" --> university_college" for the value (not the label)

src/components/dev-hub/academia-sign-up-form.js
```diff
@@ -0,0 +1,213 @@
...
+                />
+                <StyledFormSelect
+                    value={institutionType}
+                    defaultText="Institution Type"
+                    choices={institutionTypes}
+                    onChange={e => setInstitutionType(e)}
+                />
+            </InstitutionSection>
+
+            <Checkbox
```
jestapinski, line 175: Let's add `required` to this element as well! I believe we want their consent

src/components/dev-hub/academia-sign-up-form.js
```diff
@@ -0,0 +1,213 @@
...
+        ['Bootcamp', 'Bootcamp'],
+        ['Online Course', 'Online Course'],
+        ['High School', 'High School'],
+        ['University / College', 'University / College'],
+    ];
+
+    const handleSubmit = async e => {
+        e.preventDefault();
+        setCanSubmit(false);
+        const obj = {
```
jestapinski, line 88: thanks for this!

src/components/dev-hub/academia-sign-up-form.js
```diff
@@ -0,0 +1,213 @@
...
+    justify-content: space-between;
+    margin-bottom: ${size.default};
+    @media ${screenSize.upToMedium} {
+        display: block;
+    }
+`;
+
+const StyledForm = styled('form')`
+    margin: 0 auto;
+    width: 720px;
```
jestapinski, line 67: This will give us some issues when trying to use on mobile/smaller screens. Can we use something like `max-width: 720px;` and `width: 100%` to get it done?

src/components/dev-hub/academia-sign-up-form.js
```diff
@@ -0,0 +1,213 @@
...
+`;
+
+const Form = React.memo(({ setSuccess, success }) => {
```
jestapinski, line 70: Do we want to just `setSuccess` in `handleSubmit` for now to see that flow working out?

src/components/dev-hub/academia-sign-up-form.js
```diff
@@ -0,0 +1,213 @@
...
+                    }
+                />
+            </InstructorSection>
+
+            <SectionHeader>
+                <StyledSectionText>Institution's Info</StyledSectionText>
+            </SectionHeader>
+
+            <InstitutionSection>
+                <StyledInput
```
jestapinski, line 161: Can we put a `maxLength` validation on here?

src/components/dev-hub/academia-sign-up-form.js
```diff
@@ -0,0 +1,213 @@
...
+                    value={institutionName}
+                    required
+                    placeholder="Name of Institution"
+                    onChange={e => setInstitutionName(e.target.value)}
+                />
+                <StyledFormSelect
```
jestapinski, line 167: This should also be `required`

src/components/dev-hub/academia-sign-up-form.js
```diff
@@ -0,0 +1,213 @@
...
+    </SuccessContainer>
+);
+
+const AcademiaSignUpForm = () => {
+    const [success, setSuccess] = useState(null);
+    if (success) {
+        return <SuccessState />;
+    }
+    return (
+        <>
```
jestapinski, line 207:
```suggestion
```
Since we have just one element here we are returning, we don't need the Fragment `<></>`

src/components/dev-hub/academia-sign-up-form.js
```diff
@@ -0,0 +1,213 @@
...
+const StyledInput = styled(Input)`
+    height: 40px;
+    margin-bottom: ${size.large};
+    width: 335px;
+`;
+
+const StyledFormSelect = styled(FormSelect)`
+    color: ${colorMap.greyLightTwo};
+    height: 40px;
+    width: 335px;
```
jestapinski, line 26: Same issue with mobile responsiveness with fixed px widths. Can we use something like Flexbox to help provide a more responsive experience?

src/components/dev-hub/academia-sign-up-form.js
```diff
@@ -0,0 +1,200 @@
...
+    grid-template-columns: repeat(auto-fill, 335px);
+    grid-row-gap: ${size.mediumLarge};
+    justify-content: space-between;
+    margin-bottom: ${size.xlarge};
+    @media ${screenSize.upToMedium} {
+        display: block;
+    }
+`;
+
+const StyledForm = styled('form')`
```
jestapinski, line 53: We'll likely want to apply max-width and width to the general page layout as well, but we can do that separately and then remove here later.

src/components/dev-hub/academia-sign-up-form.js
```diff
@@ -0,0 +1,200 @@
...
+`;
+
+const InstructorSection = styled('div')`
+    display: grid;
+    grid-template-columns: repeat(auto-fill, 335px);
+    grid-row-gap: ${size.mediumLarge};
+    justify-content: space-between;
+    margin-bottom: ${size.xlarge};
+    @media ${screenSize.upToMedium} {
+        display: block;
```
jestapinski, line 38: the css grid is NICE! Looks 10x better already

On mobile we will need to update these styles a bit to let the grid work properly. The shape is now one column and three rows, can we update this media query to use this new layout instead of `block`? That way we will also get some nice spacing between form elements

src/components/dev-hub/academia-sign-up-form.js
```diff
@@ -0,0 +1,200 @@
...
+
+const StyledSectionText = styled(ArticleH3)`
+    color: ${colorMap.greyLightTwo};
+    font-family: akzidenz;
+    font-weight: normal;
+`;
+
+const InstructorSection = styled('div')`
+    display: grid;
+    grid-template-columns: repeat(auto-fill, 335px);
```
jestapinski, line 33: The 335 feels a little hardcoded but I get where it comes from, can we pull that out and the 720 into CONST_VARS?

src/pages/academia.js
```diff
@@ -49,6 +49,9 @@ export default () => {
         { label: 'MongoDB for Academia', target: '/academia' },
     ];
 
+    const scrollToRef = ref =>
```
jestapinski, line 52: cool! Looks like this is also [supported on all browsers](https://developer.mozilla.org/en-US/docs/Web/API/Window/scrollTo) so nice touch

src/components/dev-hub/success-state.js
```diff
@@ -0,0 +1,21 @@
...
+import SuccessIcon from './icons/success';
+
+const SuccessContainer = styled('div')`
+    text-align: center;
+    svg {
+        margin-bottom: ${size.large};
+    }
+`;
+
+const SuccessState = ({ children }) => (
```
jestapinski, line 14: Maybe I missed this but can we reference this in the project sign up form?

src/components/dev-hub/academia-sign-up-form.js
```diff
@@ -0,0 +1,200 @@
...
+`;
+
+const InstitutionSection = styled('div')`
+    display: grid;
+    grid-template-columns: repeat(auto-fill, 335px);
+    grid-row-gap: ${size.mediumLarge};
+    justify-content: space-between;
+    margin-bottom: ${size.xlarge};
+    @media ${screenSize.upToMedium} {
+        display: block;
```
jestapinski, line 49: Same grid comment

src/components/dev-hub/success-state.js
tahiya-mongo: Oops! Totally forgot to do that. Thanks!

src/components/dev-hub/academia-sign-up-form.js
```diff
@@ -0,0 +1,204 @@
...
+
+const InstitutionSection = styled('div')`
+    display: grid;
+    grid-template-columns: repeat(auto-fill, ${INPUT_BOX_WIDTH});
+    grid-row-gap: ${size.mediumLarge};
+    justify-content: space-between;
+    margin-bottom: ${size.xlarge};
+    @media ${screenSize.upToMedium} {
+        display: grid;
+        grid-template-columns: repeat(auto-fill, ${INPUT_BOX_WIDTH});
```
jestapinski, line 56:
```suggestion
        grid-template-columns: 100%;
```


Since we are using mobile, maybe just using 100% here makes sense since (after we apply margins, etc. later in the page layout) we can just flow the whole width of the screen. Setting to 335px actually will overflow the screen on some smaller devices. Here is an iphone 5 for example:

Screen Shot 2020-05-26 at 10 35 14 AM

src/components/dev-hub/academia-sign-up-form.js
```diff
@@ -0,0 +1,204 @@
...
+    }
+`;
+
+const InstitutionSection = styled('div')`
+    display: grid;
+    grid-template-columns: repeat(auto-fill, ${INPUT_BOX_WIDTH});
+    grid-row-gap: ${size.mediumLarge};
+    justify-content: space-between;
+    margin-bottom: ${size.xlarge};
+    @media ${screenSize.upToMedium} {
```
jestapinski, line 54: Looks like here and in `InstructorSection` we have css used both in and out of the media query. These styles will carry over so we can omit whatever is the same (such as `display: grid`, `grid-row-gap`, `justify-content` and `margin-bottom` all look the same so let's clean that up!)

tahiya-mongo:
Thanks!

PR #401, opened 2020-05-21
Use final dev rel YT playlist

jestapinski:
Deepak sent me [this playlist](https://www.youtube.com/playlist?list=PL4RCxklHWZ9sOT_iMTc3RKgSoVbQpe8aX) as the final devrel YouTube playlist to be used with the media component, this PR swaps to point to that one.


greggb:
Should we track this in Stitch instead?

jestapinski:
Yeah we can move this into Stitch, that is a good idea!

PR #402, opened 2020-05-21
Move playlistId logic into Stitch

jestapinski:
This PR moves the logic for which playlist ID is used for YouTube into Stitch. It also cleans up syntax so we just pass `maxResults` to Stitch and not `{maxResults}`.


jestapinski:
Thanks!

PR #404, opened 2020-05-26
Allow Staging Dev Pages

jestapinski:
This PR gives us a way to maintain pages "in development" that don't go live to prod. Simply put, if a page is "staged" then we remove it if we are building in a production environment. This allows us to continue to develop these pages locally without anything being publicly available on production.

This approach is pretty simple because we can simply case on `SNOOTY_ENV`, which is `staging` at all times unless we run a job on the workerpool. This might be a bit stricter than we want, but the ability to case on `SNOOTY_ENV` seems like the best solution here.

We also can do things like continue to build cypress tests since `SNOOTY_ENV` is `staging` for CI testing as well, so we can have testing for the correct endpoint in place before a page ships out as well.

I wrote a quick test as well for good measure, if this is good to go I will then add back in the academia page to help us continue to build that out. The media page is almost done so I'm not inclined to bring that back in.

### DIFF
--- a/src/utils/setup/on-create-page.js
+++ b/src/utils/setup/on-create-page.js
@@ -1,5 +1,6 @@
 const memoizerific = require('memoizerific');
 const { removeExcludedArticles } = require('./remove-excluded-articles');
+const { removePageIfStaged } = require('./remove-page-if-staged');
 const { getNestedValue } = require('../get-nested-value');
 const { getMetadata } = require('../get-metadata');
 
@@ -21,6 +22,8 @@ const DEFAULT_FEATURED_LEARN_SLUGS = [
     'how-to/golang-alexa-skills',
     'how-to/polymorphic-pattern',
 ];
+
+const STAGING_PAGES = ['/academia/', '/media/'];
 
 const requestStitch = async (functionName, ...args) =>
     stitchClient.callFunction(functionName, [metadata, ...args]);
@@ -171,6 +174,7 @@ const onCreatePage = async (
         default:
             break;
     }
+    removePageIfStaged(page, deletePage, STAGING_PAGES);
 };
 
 module.exports = { findArticlesFromSlugs, getLearnPageFilters, onCreatePage };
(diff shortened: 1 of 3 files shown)


src/utils/setup/remove-page-if-staged.js
```diff
@@ -0,0 +1,10 @@
+const removePageIfStaged = (page, deletePage, stagingPages) => {
+    if (
+        process.env.SNOOTY_ENV === 'production' &&
+        stagingPages.includes(page.path)
+    ) {
+        deletePage(page);
```
jestapinski, line 6: I'm assuming this also won't show on tagmanager since the page is removed before the end of this lifecycle step but not sure how to verify locally

src/utils/setup/on-create-page.js
```diff
@@ -22,6 +23,8 @@ const DEFAULT_FEATURED_LEARN_SLUGS = [
     'how-to/polymorphic-pattern',
 ];
 
+const STAGING_PAGES = ['/academia/', '/media/'];
```
greggb, line 26: I think the discoverability of this array could be an issue. Maybe we need a place in the README for 'developing'?
 
As we add more of these arrays and values it might be beneficial to have a config file we can import and update outside of the code itself.

jestapinski: I was thinking about some external source to allow these changes to not be full-on code changes. Perhaps this is a nice place for Stitch to come in?

greggb: I was thinking about that. On the one hand it would enable stakeholders to 'un-stage', but on the other it's an impediment to contributing.

Probably still the right thing to do though

jestapinski: Yeah I think there are some cons, such as adding another Stitch call to build time. I think the best course of action is to merge this PR as-is to allow us to develop on the academia page and then experiment with Stitch separately.

jestapinski:
Thanks!

PR #405, opened 2020-05-26
Academia Page Text Styling 

tahiya-mongo:
This PR adds styled text for the `Academia` page body section. The background images will be integrated once finalized.


src/pages/academia.js
```diff
@@ -42,6 +43,39 @@ const StyledButton = styled(Button)`
     margin-top: ${size.large};
 `;
 
+const BodyContent = styled('div')`
+    margin: 0 auto;
+    max-width: ${CONTENT_MAX_WIDTH};
```
jestapinski, line 48: For articles we use the max width `size.maxContentWidth` from `theme`, can we use that here? (It is going to be a similar layout)

src/pages/academia.js
```diff
@@ -42,6 +43,39 @@ const StyledButton = styled(Button)`
...
+    width: 100%;
+`;
+
+const OfferingsContent = styled('div')`
+    margin-bottom: ${size.xlarge};
+    max-width: 324px;
```
jestapinski, line 54: I have a feeling when we put images in this will change but I am fine leaving this like so for now. Likely we will use something like css grid or flexbox to handle responsiveness with the side by side layout.

src/pages/academia.js
```diff
@@ -57,22 +91,91 @@ export default () => {
...
+                    <StyledSectionHeader>For Students</StyledSectionHeader>
+                    <StyledArticleH3>
+                        MongoDB for Academia is for educators who want to
+                        prepare students for careers that require in-demand
+                        database skills that power modern applications.
+                        <br />
+                        <br />
+                        You’re eligible for this program if you teach:
+                        <StyledBullet>
+                            <li>
```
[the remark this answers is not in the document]
tahiya-mongo: I only added the `StyledSpan` tag to make sure the text color stays white. Applying the styles to `li` colors both the bullet and the text. Is there a better way to do this?

jestapinski:
Those conflicts exist because I am looking to merge the `academia-page` branch into staging so we can once again merge these page in progress PRs to staging without the page being publicly accessible. Once that is in we can swap the target branch to staging and merge staging into this branch.

src/pages/academia.js
```diff
@@ -86,12 +162,106 @@ export default () => {
                         </StyledButton>
                     </HeaderContent>
                 </Header>
-            </HeroBanner>
+            </StyledHeroBanner>
 
             <BodyContent>
-                <div ref={formRef}>
-                    <AcademiaSignUpForm />
+                <div>
```
jestapinski, line 168: So the design does a small overlap between the image and the text like so:

Screen Shot 2020-05-27 at 9 58 35 AM

You are definitely right sticking to the 780px max width, Dan's design looks to be like 850 which gives the image more space. For implementation though, let's stick to the 780px max width and shrink this image to 450 or 500px, I'll leave that up to you. That way we can give more room to the text since it was pretty narrow before:

Screen Shot 2020-05-27 at 10 03 31 AM

src/pages/academia.js
```diff
@@ -86,12 +162,106 @@ export default () => {
...
+                                        <StyledSpan>
+                                            Bootcamps and Online Courses
+                                        </StyledSpan>
+                                    </li>
+                                </StyledBullet>
+                            </BodyText>
+                        </EligibilityContent>
+                    </div>
+
+                    {!isMobile && <img src={AcademiaLeafImage} alt="" />}
```
jestapinski, line 251: Let's see if we can push this image down a bit, since it looks to belong in the For Students section

src/pages/academia.js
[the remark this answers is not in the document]
jestapinski: Ah that seems good!

src/pages/academia.js
```diff
@@ -7,48 +7,114 @@ import {
...
 
+const StyledP = styled(P)`
+    font-family: akzidenz;
+    font-size: ${fontSize.default};
+    line-height: ${lineHeight.default};
+    margin-top: ${size.articleContent};
+`;
+
+const StyledLeafImage = styled('img')`
+    margin-top: 400px;
```
jestapinski, line 115: I think this is fine since we hide the image on smaller sizes, thanks!

src/pages/academia.js
```diff
@@ -59,39 +125,136 @@ export default () => {
...
                 </div>
+                {!isMobile && (
+                    <img src={TeachMongoDBImage} alt="" width="450px" />
+                )}
+            </BodyContent>
+
+            <EligibilitySection>
+                <BodyContent>
+                    <div>
+                        <EligibilityContent>
```
jestapinski, line 198: Some stakeholders might want smaller padding around here on mobile, but I think it is worth holding off on messing with it until later.

tahiya-mongo:
Thanks for the feedback!

PR #407, opened 2020-05-26
Handle Academia Form Submits

jestapinski:
This PR adds logic for sending academia form data to a new Stitch function called `submitAcademiaRegistration`, which performs an insert into Atlas with this data. I also have a Stitch trigger setup for handling writes to Atlas (ideally to send a notification to Lieke) but this piece is incomplete.

Here we can see the data in Atlas:
Screen Shot 2020-05-26 at 2 45 53 PM

### DIFF
--- a/src/components/dev-hub/success-state.js
+++ b/src/components/dev-hub/success-state.js
@@ -11,8 +11,8 @@ const SuccessContainer = styled('div')`
     }
 `;
 
-const SuccessState = ({ children }) => (
-    <SuccessContainer>
+const SuccessState = ({ children, ...props }) => (
+    <SuccessContainer {...props}>
         <SuccessIcon />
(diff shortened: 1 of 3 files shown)


src/components/dev-hub/success-state.js
```diff
@@ -11,8 +11,8 @@ const SuccessContainer = styled('div')`
     }
 `;
 
-const SuccessState = ({ children }) => (
-    <SuccessContainer>
+const SuccessState = ({ children, ...props }) => (
+    <SuccessContainer {...props}>
```
jestapinski, line 15: I also passed `props` here to give some extra margin:

Screen Shot 2020-05-26 at 2 50 27 PM

jestapinski:
Thanks!

PR #410, opened 2020-05-27
Create new stacking context for blockquote

jestapinski:
I have found a potential solution to some of our z-index woes and have implemented it on the Blockquote.

After #347 the `GlobalWrapper` had a background which caused some unintended side-effects. One of which was hiding a piece of the `Blockquote`:

Screen Shot 2020-05-27 at 11 25 17 AM

In fact, this piece of the Blockquote had a z-index of -1. Since we did not reset the stacking context for the Blockquote, this was hiding under every parent up to `<body />`. Adding this background exposed this problem.

Screen Shot 2020-05-27 at 11 25 13 AM

This PR adds a `div` to the Blockquote which resets the stacking context, so the blockquote sits in front of other elements and using z-indices here is with respect to the blockquote only.

Screen Shot 2020-05-27 at 11 22 38 AM

Funny enough, it seems this can also help with the `Button` hover state, should we want to bring that back in.

### DIFF
--- a/src/components/dev-hub/blockquote.js
+++ b/src/components/dev-hub/blockquote.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import styled from '@emotion/styled';
 import ComponentFactory from '../ComponentFactory';
-import { colorMap, gradientMap, size } from './theme';
+import { colorMap, gradientMap, layer, size } from './theme';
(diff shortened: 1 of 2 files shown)


jestapinski:
Thanks! Yes, I feel those buttons deserve more justice than they currently have

PR #412, opened 2020-05-27
Use Emotion-Theme in Components and Pages

tahiya-mongo:
This PR adds emotion-theming to the components below to prepare for a theme change in the near future.

Components modified so far:

- article-share-footer.js
- author-image.js
-  badge.js
- blockquote.js
- blog-post-title-area.js
- blog-tag-list.js
- breadcrumb.js
- button.js
- byline-block.js
- card.js
- chart.js
- codeblock.js
- contents-menu.js
- controlled-tooltip.js
- copy-button.js
- events.js
- feature-block.js
- global-footer.js
- global-nav.js
- gradient-image.js
- hero-banner.js
- input-label.js
- input.js

### DIFF
--- a/src/components/dev-hub/article-share-footer.js
+++ b/src/components/dev-hub/article-share-footer.js
@@ -3,7 +3,7 @@ import copy from 'copy-to-clipboard';
 import styled from '@emotion/styled';
 import { getArticleShareLinks } from '../../utils/get-article-share-links';
 import BlogTagList from './blog-tag-list';
-import { colorMap, screenSize, size } from './theme';
+import { darkTheme, screenSize, size } from './theme';
 import Link from './link';
 import Tooltip from './tooltip';
 import LinkIcon from './icons/link-icon';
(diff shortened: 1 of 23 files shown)


src/components/dev-hub/article-share-footer.js
```diff
@@ -12,7 +12,7 @@ import FacebookIcon from './icons/facebook-icon';
 import TwitterIcon from './icons/twitter-icon';
 
 const ArticleShareArea = styled('div')`
-    border-top: 1px solid ${colorMap.greyDarkTwo};
+    border-top: 1px solid ${darkTheme.colorMap.greyDarkTwo};
```
jestapinski, line 15: Ah my fault, I thought the PR I linked had one or two components using the theme but alas they did not. Fun fact: we don't need to actually import and use `darkTheme` for any of these styled components.

What happens is in `layout.js` we set the theme in a component called `ThemeProvider`. This allows us to access the active theme (which may change from `darkTheme`) in our styled component declaration as such:

`border-top: 1px solid ${({theme}) => theme.colorMap.greyDarkTwo}`

Seems like magic, right? That is because when we declare the theme in `layout.js`, it gets passed to all child styled components so they can use just as any other prop.

Again sorry I thought that PR did one as such, but let's follow this pattern so the proper theme is being used, thanks!

jestapinski: You can see this in action [in the docs](https://emotion.sh/docs/theming)

tahiya-mongo: Thanks and no worries at all. I realized I didn't do it the right way after creating the PR (which is why I converted it into a draft). I'm on it!

tahiya-mongo:
Closed in order to reduce the number of components to be reviewed at once. Please refer to  [417](https://github.com/mongodb/devhub/pull/417)

PR #414, opened 2020-05-28
Use YouTube Video PublishDate for Sorting

jestapinski:
While testing the media component I found a small bug. The `publishAt` on `snippet` from YouTube is for when this video was added to the playlist. We should be sorting by when the video is published to YouTube.

I made the API call change in Stitch as well to get `contentDetails` as well, and the sorting is as expected after this PR.

[Here are the docs for playlist items](https://developers.google.com/youtube/v3/docs/playlistItems), where we can see this distinction.


jestapinski:
Thanks!

PR #415, opened 2020-05-28
Clean up prop errors in console

jestapinski:
This PR cleans up some lingering console errors related to incorrect prop casing. Specifically:
- On the nav image, we needed to camelcase in a few spots
- On cards, we conditionally render a styled component or the gatsby `Link`. If we render the latter, we must lowercase attributes as they are passed through to the DOM.


jestapinski:
Thanks!

PR #416, opened 2020-05-28
Convert Media Methods to ES5

jestapinski:
[Gatsby does not support ES5 at build time](https://github.com/gatsbyjs/gatsby/issues/7810) and there is no clean way to get ES6 in there. We would like to use these media methods at build time as well, so as a prerequisite we must use ES5 export syntax before using these at build time.

This PR simply switches the methods we would like to use to ES5 before adding them in at build time (separate PR). I can still access the Twitch video on the home page, so everything continues to work as expected.


jestapinski:
Looking into ESM

PR #418, opened 2020-05-28
Move /academia to /academia/educators

jestapinski:
This PR moves the academia page to `/academia/educators` and sets up a redirect for `/academia` to `/academia/educators`.

When running the `develop` server, there is a blip where it renders a 404 thten goes to the correct place. Fortunately, the [`gatsby-plugin-meta-redirect`](https://github.com/nsresulta/gatsby-plugin-meta-redirect) creates a small file to help handle this redirect on production builds so this will not happen after a production build.

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -43,6 +43,7 @@ module.exports = {
                 feeds: [articleRssFeed],
             },
         },
+        'gatsby-plugin-meta-redirect', // this must be last
     ],
     siteMetadata: {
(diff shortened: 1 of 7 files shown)


jestapinski:
> Nice. How does the interaction with the staging_pages and redirect work?

Good question! The redirect occurs and then `/academia/educators` gives a 404.

jestapinski:
Thanks!

PR #419, opened 2020-05-29
Use "esm" for ECMAScript Imports at Buildtime

jestapinski:
This PR uses a package called [`esm`](https://github.com/standard-things/esm) to allow us to use more modern import syntax at buildtime. This was surprisingly painless to add!


gatsby-node.esm.js
```diff
@@ -0,0 +1,168 @@
+const path = require('path');
```
jestapinski, line 1: There are no code differences between here at the previous `gatsby-node.js`. I just moved all of that code into this file for `esm`.

gatsby-node.js
```diff
@@ -1,168 +1,2 @@
...
-    onCreatePage(
-        page,
-        actions,
-        stitchClient,
-        homeFeaturedArticles,
-        learnFeaturedArticles,
-        excludedLearnPageArticles
-    );
+require = require('esm')(module);
+module.exports = require('./gatsby-node.esm.js');
```
jestapinski, line 2: Here is the `esm` magic, suggested in [this Gatsby issue thread](https://github.com/gatsbyjs/gatsby/issues/7810#issuecomment-449741977)

gatsby-node.esm.js
greggb: In a future PR we can move all the `require`s and `exports` to es syntax 🎉

jestapinski:
Thanks! Yes I plan to address the es syntax shortly

PR #420, opened 2020-05-29
Convert ES5 require syntax to import/export for gatsby-node

jestapinski:
This PR follows off the esm work to convert gatsby-node related modules to use `import/export` syntax. No functionality has been added or removed.

I renamed the helper module `onCreatePage` to `handleCreatePage` since this now conflicts with the gatsby node API function name. Before with `exports.onCreateNode` this was not an issue.


jestapinski:
Thanks! It does build and everything seems fine

PR #421, opened 2020-06-01
Update form error handling UX

jestapinski:
This PR changes a few things on the academia form, and one similar change on the project sign up form:
- Fixed a bug where the Checkbox required message did not show on Chrome (thinking about opening a PR for Leafygreen for this, since it seems important enough)
- Error messages for first name, last name, and email now show a "cannot be blank" message if blank instead of just the type validation
- Removed validation regex for email (also done on project sign up form)

Screen Shot 2020-06-01 at 11 43 52 AM

### DIFF
--- a/src/components/dev-hub/academia-sign-up-form.js
+++ b/src/components/dev-hub/academia-sign-up-form.js
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useCallback, useState } from 'react';
 import styled from '@emotion/styled';
 import { submitAcademiaForm } from '../../utils/devhub-api-stitch';
 import { colorMap, size, screenSize } from './theme';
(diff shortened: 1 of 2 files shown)


src/components/dev-hub/academia-sign-up-form.js
```diff
@@ -17,6 +17,14 @@ const StyledButton = styled(Button)`
     margin: ${size.large} auto;
 `;
 
+const StyledCheckbox = styled(Checkbox)`
+    /* Leafygreen's Checkbox input is given height/width of 0, hiding invalid popup */
```
jestapinski, line 21: 🤷‍♂️  looks like this element was actually hidden with height/width 0. Chrome's error message dialog likely could not bind to it. Still functional in other browsers with this change.

greggb: Does this still show the custom checkbox? They probably did this to use the events on the checkbox, but to display a totally custom implementation.

jestapinski: Yes, this still does!

jestapinski: The `input` sits behind the custom element, so the behavior still works as intended

jestapinski:
Thanks!

PR #422, opened 2020-06-01
Access devhub stitch functions at build time

jestapinski:
When we added our new Stitch functions into the app, we used a library called `mongodb-stitch-browser-sdk`. This was appropriate because these functions were only being called by the client at runtime.

However, by moving these calls into buildtime we can no longer reference `mongodb-stitch-browser-sdk` and must use `mongodb-stitch-server-sdk` (as DOCSP was already doing at buildtime with fetching from the Snooty app). In fact, if at buildtime we imported from any module that would import from `mongodb-stitch-browser-sdk` we would get a buildtime error.

This PR attempts to split logic to allow for us to use some of the existing logic for adding youtube and twitch videos at buildtime. We still have to use `mongodb-stitch-server-sdk` so I mimiced the process we were already using for getting articles.


src/utils/fetch-youtube-data.js
```diff
@@ -1,27 +1,12 @@
-import dlv from 'dlv';
 import { requestYoutubePlaylist } from './devhub-api-stitch';
+import { simplifyYoutubeResponse } from './simplify-youtube-response';
 // Fetches data from youtube api
 
-const simplifyResponse = responseData => {
```
jestapinski, line 5: These mappings needed to be pulled out of these modules entirely, since they would import from a module that uses `mongodb-stitch-browser-sdk`.

src/utils/setup/init-stitch.js
```diff
@@ -1,10 +1,10 @@
 import { Stitch, AnonymousCredential } from 'mongodb-stitch-server-sdk';
 import { SNOOTY_STITCH_ID } from '../../build-constants';
 
-export const initStitch = async () => {
-    const stitchClient = Stitch.hasAppClient(SNOOTY_STITCH_ID)
-        ? Stitch.getAppClient(SNOOTY_STITCH_ID)
-        : Stitch.initializeAppClient(SNOOTY_STITCH_ID);
+export const initStitch = async (id = SNOOTY_STITCH_ID) => {
```
jestapinski, line 4: Since this module only imports from `mongodb-stitch-server-sdk`, we can continue to use it. I parameterized the `id` here to allow it to work for both Stitch apps at build-time

src/utils/fetch-youtube-data.js
jestapinski: They can (and maybe should) move into Stitch as well, seems to be more server-side logic. Thoughts?

greggb: It would keep our end use of the API data consistent, but we'd just need to make sure it's noted that if the API from Twitch changes we need to update at the stitch level.

jestapinski: Yeah I think I will tackle that separately

jestapinski:
Thanks!

PR #423, opened 2020-06-01
Podcasts in Stitch, Build Time

jestapinski:
This PR moves the `fetch` for Lybsin Podcasts also into Stitch. `fetch` is not supported at build-time since `window` is undefined 🤦  so this is a requirement to use podcasts at build-time. Part of this was moving the postprocessing `parsePodcasts` into its own module.

I also updated the function to get build time data to support both videos and podcasts, to save on overhead for instantiating a stitch client.

This is the last piece of work to get media elements pulled at build time.


jestapinski:
Thanks all

PR #428, opened 2020-06-02
Use build-time media on the learn page

jestapinski:
This PR builds off of the previous work to move media fetching into build-time to use this media data on the learn page. This removes the "jittering" effect so long as there is no new content between the latest build and when the page is rendered.

I updated the hooks to take a default value, so that way if the fetch returns identical data there should not be a rerender.

The target branch for this PR is not staging, it is the branch with media on the learn page.

### DIFF
--- a/src/utils/setup/fetch-build-time-media.js
+++ b/src/utils/setup/fetch-build-time-media.js
@@ -14,10 +14,10 @@ export const fetchBuildTimeMedia = async () => {
         client.callFunction('fetchLybsinPodcasts', []),
     ]);
     return {
-        videos: [
+        allVideos: [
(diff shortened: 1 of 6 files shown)


src/utils/setup/fetch-build-time-media.js
```diff
@@ -14,10 +14,10 @@ export const fetchBuildTimeMedia = async () => {
         client.callFunction('fetchLybsinPodcasts', []),
     ]);
     return {
-        videos: [
+        allVideos: [
```
jestapinski, line 17: Renamed these to avoid naming collisions

jestapinski:
Thanks!

PR #430, opened 2020-06-02
Academia page design review updates

tahiya-mongo:
This PR replaces the Hero banner image with the new image provided by Dan and increases the checkbox text size to `16px`.


src/images/1x/Academia.svg
```diff
@@ -0,0 +1,133 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jestapinski, line 1: Yeah this new image looks much better. Something about it seems odd on mobile but I'll bring this up with Dan. Thanks!

Screen Shot 2020-06-02 at 4 04 47 PM

src/components/dev-hub/academia-sign-up-form.js
```diff
@@ -23,6 +23,9 @@ const StyledCheckbox = styled(Checkbox)`
         height: 20px;
         width: 20px;
     }
+    > span {
+        font-size: 16px;
```
jestapinski, line 27: Can we use a value from `theme` here? Above I didn't because the checkbox is specifically 20px (and won't change)

tahiya-mongo: Yep! I meant to go back and change that but totally forgot

PR #432, opened 2020-06-03
Move logic for staged pages into Stitch

jestapinski:
This PR moves the local variable `STAGING_PAGES` into Stitch and adds support to access it. This will allow us to control when pages are released without needing a code change on the repo (so long as the change is done before a new build).

In Stitch I set up a basic value called `STAGING_PAGES` which is just an array of pages we don't want out on Prod yet. I also set up a very simple function called `fetchStagingPages` which serves as a getter for this value.

### DIFF
--- a/src/utils/setup/handle-create-page.js
+++ b/src/utils/setup/handle-create-page.js
@@ -4,6 +4,7 @@ import { removePageIfStaged } from './remove-page-if-staged';
 import { getNestedValue } from '../get-nested-value';
 import { getMetadata } from '../get-metadata';
 import { fetchBuildTimeMedia } from './fetch-build-time-media';
+import { getStagingPages } from './get-staging-pages';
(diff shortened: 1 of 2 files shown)


src/utils/setup/handle-create-page.js
```diff
@@ -24,7 +25,9 @@ const DEFAULT_FEATURED_LEARN_SLUGS = [
     'how-to/polymorphic-pattern',
 ];
 
-const STAGING_PAGES = ['/academia/educators/', '/media/'];
+const memoizedStagingPages = memoizerific(1)(
```
jestapinski, line 28: I memoized this call since we would otherwise request Stitch once for each page that is built to get the Staging Pages. Now we only make one request.

jestapinski:
Thanks!

PR #435, opened 2020-06-03
Swap Nav link to Community Forums

jestapinski:
[DEVHUB-150](https://jira.mongodb.org/browse/DEVHUB-150)

This PR switches the "community" link on the top nav to point to the community forums instead of the existing community page.

This PR does not touch the community page files, since it seems like DevRel will update these in the future. They still live and are accessible at `/community` and `/community/events`. We can stage these, but not sure what the best course of action here is.


jestapinski:
Thanks! I will merge this in after the academia deploy

PR #436, opened 2020-06-03
Add flexbox constraints for Academia Images

jestapinski:
This PR addresses images being different sizes on various browsers on the academia page.

The issue was Chrome was reading the `width` attribute off of images to provide flexbox positioning. Firefox did not work the same way. The solution here was to use flexbox css to specify an exact pixel width for these images.

Now the images look great on any browser. I also moved the leaf one up a bit to get it in line with the design.

Before (FF):
Screen Shot 2020-06-03 at 3 23 18 PM
Screen Shot 2020-06-03 at 3 23 22 PM

After (FF):
Screen Shot 2020-06-03 at 3 17 47 PM
Screen Shot 2020-06-03 at 3 26 14 PM


jestapinski:
Thanks!

PR #437, opened 2020-06-03
Code for Good page 

tahiya-mongo:
This is PR adds the new `Code for Good` page for DevHub.

### DIFF
--- a/src/pages/code-for-good.js
+++ b/src/pages/code-for-good.js
@@ -17,6 +17,7 @@ import BackgroundWaves from '../images/code-for-good/background-waves.png';
 import WildAidLogo from '../images/code-for-good/wild-aid-logo.png';
 import WildAidDashboard from '../images/code-for-good/wild-aid-dashboard.png';
 import VesselRecord from '../images/code-for-good/vessel-record.png';
+import Link from '../components/dev-hub/link';
 
 const CAPTION_MAX_WIDTH = '345px';
 const BODY_TEXT_MAX_WIDTH = '320px';
@@ -31,7 +32,7 @@ const Header = styled('header')`
     justify-content: space-between;
     padding-top: ${size.large};
     @media ${screenSize.upToLarge} {
-        display: block;
+        flex-direction: column-reverse;
         margin-bottom: ${size.large};
     }
 `;
@@ -40,14 +41,16 @@ const HeaderText = styled('div')`
     max-width: ${HERO_TEXT_WIDTH};
 `;
 
-const HeaderLink = styled('a')`
+const HeaderLink = styled(Link)`
     color: ${colorMap.darkGreen};
+    display: inline;
     margin-top: ${size.large};
     margin-left: ${size.mediumLarge};
-    text-decoration: none;
+    white-space: nowrap;
     @media ${screenSize.upToLarge} {
         display: block;
-        margin-left: auto;
+        font-size: ${fontSize.xsmall};
+        margin: 0;
     }
 `;
 
@@ -61,7 +64,6 @@ const StyledHeroBanner = styled(HeroBanner)`
 `;
 
 const StyledP = styled(P)`
-    font-family: akzidenz;
     font-size: ${fontSize.default};
     line-height: ${lineHeight.default};
     margin-top: ${size.articleContent};
@@ -69,7 +71,6 @@ const StyledP = styled(P)`
 `;
 
 const BodyText = styled(P)`
-    font-family: akzidenz;
     font-size: ${fontSize.default};
     line-height: ${lineHeight.default};
     margin-top: ${fontSize.default};
@@ -77,9 +78,8 @@ const BodyText = styled(P)`
 
 const StyledButton = styled(Button)`
     margin-top: ${size.large};
-    margin-bottom: ${size.xlarge};
     @media ${screenSize.upToLarge} {
-        margin-bottom: ${size.small};
+        margin-bottom: ${size.medium};
     }
 `;
 
@@ -92,24 +92,25 @@ const BodyContent = styled('div')`
     }
 `;
 
-const StyledLink = styled('a')`
+const StyledLink = styled(Link)`
     color: ${colorMap.darkGreen};
+    font-weight: bold;
     margin-top: ${size.mediumLarge};
     text-decoration: none;
 `;
 
-const FirstSection = styled('div')`
+const UnderstandTextSection = styled('div')`
     margin-bottom: ${size.xlarge};
     margin-top: ${size.xlarge};
     max-width: ${BODY_TEXT_MAX_WIDTH};
 `;
 
-const SecondSection = styled('div')`
+const VesselAppTextContainer = styled('div')`
     margin-bottom: ${size.xlarge};
     max-width: ${BODY_TEXT_MAX_WIDTH};
 `;
 
-const StyledDiv = styled('div')`
+const VesselAppContainer = styled('div')`
     display: flex;
     justify-content: space-between;
     @media ${screenSize.upToMedium} {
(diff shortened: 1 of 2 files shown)


src/pages/code-for-good.js
```diff
@@ -0,0 +1,304 @@
...
+                        </SecondSection>
+                    </div>
+
+                    <VesselImageDiv>
+                        <StyledVesselImage
+                            src={VesselRecord}
+                            alt=""
+                            height="614px"
+                            width="345px"
+                        />
```
jestapinski, line 293: We can remove these values now (I think?) since they are in the css

src/pages/code-for-good.js
```diff
@@ -0,0 +1,304 @@
...
+    line-height: ${lineHeight.default};
+    margin-top: ${fontSize.default};
+`;
+
+const StyledButton = styled(Button)`
+    margin-top: ${size.large};
+    margin-bottom: ${size.xlarge};
+    @media ${screenSize.upToLarge} {
+        margin-bottom: ${size.small};
+    }
```
jestapinski, line 83: This should be on the Header and not on the Button, since the Header is controlling the layout for its children (that will help some odd spacing issues with the link on smaller devices)

src/pages/code-for-good.js
```diff
@@ -0,0 +1,285 @@
...
+                    </div>
+
+                    <StyledVesselImage>
+                        <img
+                            src={VesselRecord}
+                            alt=""
+                            height="614px"
+                            width="345px"
+                        />
+                        <VesselCaptionText>
```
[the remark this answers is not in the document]
tahiya-mongo: Thanks for introducing me to this tag!

src/pages/code-for-good.js
```diff
@@ -0,0 +1,320 @@
...
+
+const HeaderLink = styled(Link)`
+    color: ${colorMap.darkGreen};
+    display: inline;
+    margin-top: ${size.large};
+    margin-left: ${size.mediumLarge};
+    white-space: nowrap;
+    @media ${screenSize.upToLarge} {
+        display: block;
+        font-size: ${fontSize.xsmall};
```
jestapinski, line 52: Let's make this 14px to match the button above please!

src/pages/code-for-good.js
```diff
@@ -0,0 +1,320 @@
...
+    margin-bottom: -${size.xlarge};
+    max-width: ${HERO_TEXT_WIDTH};
+`;
+
+const HeaderLink = styled(Link)`
+    color: ${colorMap.darkGreen};
+    display: inline;
+    margin-top: ${size.large};
+    margin-left: ${size.mediumLarge};
+    white-space: nowrap;
```
jestapinski, line 49: Nice!

src/pages/code-for-good.js
```diff
@@ -0,0 +1,320 @@
...
+    @media ${screenSize.upToLarge} {
+        display: block;
+        font-size: ${fontSize.xsmall};
+        margin: 0;
+    }
+`;
+
+const Title = styled(ArticleH2)`
```
jestapinski, line 57: Can this just be `ArticleH4`? That uses the smaller font and the correct font family is applied there as well

src/pages/code-for-good.js
```diff
@@ -0,0 +1,320 @@
...
+    font-size: ${fontSize.small};
+    font-family: 'Fira Mono';
+`;
+
+const StyledHeroBanner = styled(HeroBanner)`
+    margin: 0 ${size.medium};
```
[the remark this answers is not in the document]
tahiya-mongo: I agree! Looks much better

src/pages/code-for-good.js
```diff
@@ -0,0 +1,320 @@
...
+                            <ArticleH2>
+                                Contribute code to help protect our oceans
+                            </ArticleH2>
+                            <BodyText>
+                                Submit new features, translations, and styles to
+                                O-FISH. This is an open source application;
+                                anyone can contribute.
+                            </BodyText>
+
+                            <StyledLink href="#">
```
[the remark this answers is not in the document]
tahiya-mongo: The only reason some of these links aren't colored is beacuse we don't have a link yet. The color should change once we replace the `#` with an actual link.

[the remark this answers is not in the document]
jestapinski: Ah! Alright thanks!

src/pages/code-for-good.js
```diff
@@ -0,0 +1,317 @@
...
+const BodyContent = styled('div')`
+    margin: 0 auto;
+    max-width: ${size.maxContentWidth};
+    width: 100%;
+    @media ${screenSize.upToMedium} {
+        padding: 0 ${size.default};
+    }
+`;
+
+const StyledParagraphLink = styled('a')`
```
tahiya-mongo, line 88: Had to use `<a/>` here instead of `Link` because the text color wasn't showing up for some reason (on both chrome and firefox)

jestapinski: Hmm this should definitely be a `Link`. Check out the `Link` styles, is there somewhere we can override color here? We should be able to achieve that with CSS

src/pages/code-for-good.js
```diff
@@ -0,0 +1,317 @@
...
+                    <VesselImageDiv>
+                        <StyledVesselImage
+                            src={VesselRecord}
+                            alt="Sample O-Fish application with vessel boarding information"
+                        />
+                        <VesselCaptionText>
+                            Realm Sync enables officers to see and submit
+                            boarding reports through the mobile app even when
+                            they are out at sea, offline.
+                        </VesselCaptionText>
```
jestapinski, line 311: Thanks for looking at this! Let's make this `figcaption` child a `P3` like we do in articles to distinguish this text from the text below it (and play around with the color a bit). I think the design should follow our standards for `figure` from articles. One such article you can see this in is /quickstart/java-setup-crud-operations

This is what we do in articles:
```
const StyledCaption = styled(P3)`
    color: ${colorMap.greyLightTwo};
`;
const Caption = StyledCaption.withComponent('figcaption');
```

src/pages/code-for-good.js
```diff
@@ -0,0 +1,317 @@
...
+                                O-FISH. This is an open source application;
+                                anyone can contribute.
+                            </BodyText>
+
+                            <StyledLink href="#">
+                                Contribute your code
+                            </StyledLink>
+                        </VesselAppTextContainer>
+
+                        <VesselAppTextContainer>
```
jestapinski, line 286: Lets have all of the text containers go full width on mobile, no need to constrain them to the max width at that point

greggb:
@tahiya-mongo @jestapinski Once this looks good design-wise let's stage with the missing pieces to share with POs

jestapinski:
> @tahiya-mongo @jestapinski Once this looks good design-wise let's stage with the missing pieces to share with POs

Sounds good, I will set this up when it is ready

src/pages/code-for-good.js
```diff
@@ -0,0 +1,327 @@
...
+                            <StyledP>
+                                Use your current MongoDB skills and pick up new
+                                ones. Contribute back to O-FISH and help protect
+                                our oceans.
+                            </StyledP>
+                        </HeaderText>
+                        <StyledButton
+                            primary
+                            hasArrow={false}
+                            to="https://o-fish.github.io"
```
jestapinski, line 217: ah! So with these links we have to use `href` and not `to`. The difference is with `to` it is a link to somewhere on the DevHub (or same-app) and `href` can be anywhere. Let's swap these external links to href!

src/pages/code-for-good.js
```diff
@@ -0,0 +1,327 @@
...
+    @media ${screenSize.upToMedium} {
+        width: 100%;
+    }
+`;
+
+const VesselAppTextContainer = styled('div')`
+    margin-bottom: ${size.xlarge};
+    max-width: ${BODY_TEXT_MAX_WIDTH};
+    @media ${screenSize.upToMedium} {
+        width: 100%;
```
jestapinski, line 116: I think for firefox and safari we also need to set max-width to `100%`  as well here

PR #438, opened 2020-06-04
Add alt tags to images on academia page

jestapinski:
We also can add `alt` tags to academia page images to help with screen reading. This PR adds alt tags to the two body content images on the academia page.

Feedback on content welcome!


jestapinski:
Thanks!

PR #446, opened 2020-06-05
Set CFG visited links to green using href

jestapinski:
These would show as white when `visited`, but we would always like them as green.

Screen Shot 2020-06-05 at 10 08 38 AM


jestapinski:
Thanks!

PR #447, opened 2020-06-05
Fix text sizing, background waves cover

jestapinski:
This PR addresses some design tweaks for the Code for Good page. Specifically, this PR:
- Disables the `shouldContainBackground` prop on the hero banner, allowing the image to break aspect ratio while shrinking (keeping the waves grounded to the bottom of the banner).
- Disables the pink hover state on the hero banner link
- Uses flexbox to manage when to wrap the hero banner link and gives spacing between the button and link when wrapped.

Here are some screenshots:

Screen Shot 2020-06-05 at 11 47 20 AM
Screen Shot 2020-06-05 at 11 47 48 AM

### DIFF
--- a/src/pages/code-for-good.js
+++ b/src/pages/code-for-good.js
@@ -31,8 +31,8 @@ const Header = styled('header')`
     justify-content: space-between;
     padding-top: ${size.large};
     @media ${screenSize.upToLarge} {
-        flex-direction: column-reverse;
         margin-bottom: ${size.large};
+        flex-direction: column-reverse;
     }
 `;
 const HeaderText = styled('div')`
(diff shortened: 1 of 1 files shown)


src/pages/code-for-good.js
```diff
@@ -183,6 +181,13 @@ const StyledDashboardCaption = styled(P3)`
 
 const DashboardCaption = StyledDashboardCaption.withComponent('figcaption');
 
+const HeaderActionsContainer = styled('div')`
+    display: flex;
+    flex-wrap: wrap;
```
jestapinski, line 186: Using `flex-wrap` allows us to send the link to the next line instead of wrapping (nearly the same as using white-space: nowrap), but we also get more control over alignment with flexbox (such as centering elements so text aligns).

src/pages/code-for-good.js
```diff
@@ -183,6 +181,13 @@ const StyledDashboardCaption = styled(P3)`
...
+    align-items: center;
+    padding-top: ${size.large};
```
jestapinski, line 188: This also allows us to put a basic space between the actions and the above element, then add a bit more which gives spacing between elements when the wrap happens.

jestapinski:
Thanks!

PR #456, opened 2020-06-09
Emotion Theme for Component Batch A-C

jestapinski:
For merging in component changes with emotion theme, we are going to batch changes into smaller PRs to help us focus on each change to make sure no functionality changes.

This PR batches components with names starting A-C (17 components) to swap to emotion theme.

For this PR I ran a job to [create a shareable link of these changes](http://docs-devhub-staging.s3-website-us-east-1.amazonaws.com/f2c1ac5/devhub/docsworker-xlarge/emotion-theme-ac/) for QA.

PR #457, opened 2020-06-09
Emotion Theme for Component Batch E-I

jestapinski:
Similar to #456 but this PR batches components starting with E-I. Most of the ones here are icons.

I similarly ran a job for this batch, and it is viewable [here](http://docs-devhub-staging.s3-website-us-east-1.amazonaws.com/9ad03c3/devhub/docsworker-xlarge/emotion-themeing-ei/). 

I'll check this one in the morning.


jestapinski:
I just went through and gave these a look. Found one bug with the event text color so I pushed up a change for that. Otherwise this should be good to go!

tahiya-mongo:
> I just went through and gave these a look. Found one bug with the event text color so I pushed up a change for that. Otherwise this should be good to go!

Nice catch!

jestapinski:
Thanks!

PR #461, opened 2020-06-11
Emotion Theme for Component Batch L-U

jestapinski:
This is the last batch of components to be swapped to emotion theme (we still have to do the pages as well, but that should be short).

One note is the link snapshots had to be updated and we had to wrap our tests in the theme as well when mounting.

Like the other batches, I ran a job for this. [Here is the link to this branch built](http://docs-devhub-staging.s3-website-us-east-1.amazonaws.com/c70c8c3/devhub/docsworker-xlarge/emotion-theme-lu/).

### DIFF
--- a/src/components/dev-hub/link.js
+++ b/src/components/dev-hub/link.js
@@ -2,7 +2,8 @@ import React from 'react';
 import { css } from '@emotion/core';
 import styled from '@emotion/styled';
 import { Link as RouterLink } from 'gatsby';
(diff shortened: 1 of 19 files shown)


jestapinski:
I didn't take a good look yet so not sure!

jestapinski:
Just took a look and I believe this is good to go in

jestapinski:
Thanks!

PR #466, opened 2020-06-26
Delete CODEOWNERS

greggb:
This was helpful when we were developing more frequently, but I think we can manually assign reviewers now. Open to feedback though.


greggb:
Thanks!

PR #470, opened 2020-07-09
[DEVHUB-84] Comma-split Three or More Authors

jestapinski:
[Ticket](https://jira.mongodb.org/browse/DEVHUB-84)

This PR updates the BylineBlock logic to comma-split lists of authors when three or more are present. One example of this locally is in the article `/article/johns-hopkins-university-covid-19-data-atlas` which has four authors. I decided to abstract this logic out and test it since it got a little complicated.

- If there is one author, it should look like 'By Author Name' (same as before)
- If exactly two: 'By First Author and Second Author' (same as before)
- If 3+: 'By First Author, Second Author, and Third Author' (new!)

(The tests should help shed light on this)

Before:
Screen Shot 2020-07-09 at 5 47 07 PM

After:
Screen Shot 2020-07-09 at 5 44 26 PM

### DIFF
--- a/src/utils/get-author-prefix.js
+++ b/src/utils/get-author-prefix.js
@@ -1,0 +1,17 @@
+/**
+ * Generates a prefix for an author's name on Byline Block based on index and
+ * total number of authors
+ */
+export const getAuthorPrefix = (index, numberOfAuthors) => {
+    switch (index) {
+        case 0:
+            return 'By ';
+        case numberOfAuthors - 1:
+            // Handle case of two authors, so no comma needed
+            if (numberOfAuthors === 2) return '\u00a0and ';
+            return ', and ';
+        default:
+            // Otherwise, let's use commas to separate authors
+            return ', ';
+    }
+};
--- a/tests/utils/get-author-prefix.test.js
+++ b/tests/utils/get-author-prefix.test.js
@@ -1,0 +1,20 @@
+import { getAuthorPrefix } from '../../src/utils/get-author-prefix';
+
+const getFullBylineString = arr =>
+    arr.map((name, i) => `${getAuthorPrefix(i, arr.length)}${name}`).join('');
+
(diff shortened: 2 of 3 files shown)


src/utils/get-author-prefix.js
```diff
@@ -0,0 +1,17 @@
...
+ * Generates a prefix for an author's name on Byline Block based on index and
+ * total number of authors
+ */
+export const getAuthorPrefix = (index, numberOfAuthors) => {
+    switch (index) {
+        case 0:
+            return 'By ';
+        case numberOfAuthors - 1:
+            // Handle case of two authors, so no comma needed
+            if (numberOfAuthors === 2) return '\u00a0and ';
```
jestapinski, line 11: We need the unicode here to preserve a leading space. We also had this before.

tests/utils/get-author-prefix.test.js
```diff
@@ -0,0 +1,20 @@
...
+    const singleAuthorResult = getFullBylineString(singleAuthorArray);
+    expect(singleAuthorResult).toBe('By Single Author');
+
+    const twoAuthorArray = ['First Author', 'Second Author'];
+    const twoAuthorResult = getFullBylineString(twoAuthorArray);
+    expect(twoAuthorResult).toBe('By First Author\u00a0and Second Author');
+
+    const threeAuthorArray = ['First Author', 'Second Author', 'Third Author'];
+    const threeAuthorResult = getFullBylineString(threeAuthorArray);
+    expect(threeAuthorResult).toBe(
```
jbouzi12, line 17: Small nit, should we add one more example of a long list, like 5 or more authors? (is that a possible scenario?).

jestapinski: You can never be too careful! I added a test for 6

jestapinski:
Thanks!

PR #474, opened 2020-08-06
Fix local URLs coming from Events API

jestapinski:
[Staging Link](https://docs-mongodbcom-staging.corp.mongodb.com/master/devhub/jordanstapinski/js/fix-event-local-urls/community/events/)
[Broken Production Link](https://developer.mongodb.com/community/events/)

This PR fixes addresses an issue where event data is coming in with relative links to an event on mongodb.com instead of absolute links. We don't try to format them as-is so this would try to link to developer.mongodb.com/community/events/events/eventname instead of mongodb.com/events/eventname

The solution here was to regex check for either `http` or `www` starting off this field. If this is the case then we don't add anything to the URL. If not, we add the mongodb.com/ prefix to make the link absolute.

### DIFF
--- a/src/utils/fetch-event-data.js
+++ b/src/utils/fetch-event-data.js
@@ -1,5 +1,6 @@
-export const EVENTS_URL =
+const EVENTS_API_URL =
     'https://www.mongodb.com/api/event/all/1?sort=-node_type_attributes.event_start';
+const EVENTS_BASE_URL = 'http://www.mongodb.com/';
 
 export const removePastEvents = events => {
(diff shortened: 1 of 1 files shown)


src/utils/fetch-event-data.js
```diff
@@ -1,5 +1,6 @@
-export const EVENTS_URL =
+const EVENTS_API_URL =
```
jestapinski, line 1: Checked and this was not being imported anywhere else

jestapinski:
Thank you!

PR #478, opened 2020-08-17
Remove Share Links Trailing Slash

jestapinski:
[JIRA](https://jira.mongodb.org/browse/DEVHUB-200)
[Staging Link (Sample Article)](https://docs-mongodbcom-staging.corp.mongodb.com/master/devhub/jordanstapinski/js/share-trailing-slash/how-to/static-website-deployments-mongodb-stitch-hugo-git-travis-ci)

This PR removes trailing slashes from our social share links on articles, including the copy functionality, twitter, facebook, and linkedin* (see below). I also was going to refactor these two components a bit more but it seemed pretty large for just this small change so I decided to hold off.

To verify:
- Check out the sample article
- Share links should all have the article url without a trailing slash (LinkedIn is odd and keeps putting it there even though we don't pass it, will look into this separately)
- Share links on top and bottom should be identical

LinkedIn is still putting a trailing slash on because the server seems to redirect it to do so. This is something for the backend team to look into.

### DIFF
--- a/cypress/integration/article.js
+++ b/cypress/integration/article.js
@@ -1,5 +1,5 @@
 const ARTICLE_WITH_SERIES_URL =
-    '/article/3-things-to-know-switch-from-sql-mongodb/';
+    '/article/3-things-to-know-switch-from-sql-mongodb';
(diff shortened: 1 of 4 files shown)


jestapinski:
Thanks!

PR #480, opened 2020-08-20
DEVHUB-190: Correctly use og description where appropriate

jestapinski:
[JIRA](https://jira.mongodb.org/browse/DEVHUB-190)
[Staging (Article with og description in addition to meta description)](https://docs-mongodbcom-staging.corp.mongodb.com/master/devhub/jordanstapinski/DEVHUB-190/article/3-things-to-know-switch-from-sql-mongodb)
[Staging (Article with just a meta description)](https://docs-mongodbcom-staging.corp.mongodb.com/master/devhub/jordanstapinski/DEVHUB-190/article/active-active-application-architectures)

This PR incorporates an `og:description` for SEO on an article where appropriate. The rules are as follows:
- For the `og:description` tag, we use the og description if the article provides it. If the article does not provide an og description, we fallback to the meta description.
- For the meta `description` tag, we use the meta description if the article provides it. If the article does not provide this field, we fallback with the og description
- If neither are provided (very unlikely to happen), neither tag will render

The top staging link is for an article I updated to have a distince `og:description`. You can check the page source and see the distinction as such (and in the cypress tests!):
Screen Shot 2020-08-20 at 10 48 05 AM

The bottom link is for an article without an og:description provided. The `og:description` and meta `description` tags have identical content here.

### DIFF
--- a/src/components/meta-description.js
+++ b/src/components/meta-description.js
@@ -1,9 +1,0 @@
-import React from 'react';
-import { Helmet } from 'react-helmet';
-import { getNestedText } from '../utils/get-nested-text';
-
-export default ({ nodeData: { children } }) => (
-    <Helmet>
-        <meta property="og:description" content={getNestedText(children)} />
-    </Helmet>
-);
--- a/cypress/integration/article.js
+++ b/cypress/integration/article.js
@@ -2,9 +2,16 @@ const ARTICLE_WITH_SERIES_URL =
     '/article/3-things-to-know-switch-from-sql-mongodb';
 const PROD_ARTICLE_URL = `https://developer.mongodb.com${ARTICLE_WITH_SERIES_URL}`;
 
+// Article with no og description (test meta description fallback)
+const ARTICLE_WITHOUT_OG_DESCRIPTION_URL =
+    '/article/active-active-application-architectures';
+const ARTICLE_WITHOUT_OG_META_DESCRIPTION =
+    'This post will begin by describing the database capabilities required by modern multi-data center applications.';
+
 const ARTICLE_TITLE = '3 Things to Know When You Switch from SQL to MongoDB';
 const ARTICLE_DESCRIPTION =
     'Discover the 3 things you need to know when you switch from SQL to MongoDB';
+const OG_DESCRIPTION = 'og-description text';
 const SERIES_TITLE = 'SQL to MongoDB';
 
 // Images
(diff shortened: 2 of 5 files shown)


src/components/meta-description.js
```diff
@@ -4,6 +4,6 @@ import { getNestedText } from '../utils/get-nested-text';
 
 export default ({ nodeData: { children } }) => (
     <Helmet>
-        <meta property="og:description" content={getNestedText(children)} />
+        <meta name="description" content={getNestedText(children)} />
```
jestapinski, line 7: I was going to remove this component all together, but realized it could be used elsewhere since it is in the `ComponentFactory`. With that said I cleaned this up to be the meta description, not the og description.

jestapinski:
> Don't know a whole lot about meta tags. But looking through the page source for both article links and the code, it lgtm.

Ah! So here, the meta tags provide some SEO and social share value which get some cool effects. The `<meta name="description" content=<some content>/>` tag for example provides search engines with a short description to be featured about the page (what you see on Google below the url title)

`og` stands for `open-graph` as in the [open graph protocol](https://ogp.me/). These are used by various social media platforms to generate previews, etc. about a page when a link is shared (this is how some nice image previews are done) with its own set of tags, such as `og:url` `og:description`, and `og:image`

tahiya-mongo:
> > Don't know a whole lot about meta tags. But looking through the page source for both article links and the code, it lgtm.
> 
> Ah! So here, the meta tags provide some SEO and social share value which get some cool effects. The `<meta name="description" content=<some content>/>` tag for example provides search engines with a short description to be featured about the page (what you see on Google below the url title)
> 
> `og` stands for `open-graph` as in the [open graph protocol](https://ogp.me/). These are used by various social media platforms to generate previews, etc. about a page when a link is shared (this is how some nice image previews are done) with its own set of tags, such as `og:url` `og:description`, and `og:image`

Thanks for the explanation! I looked them up a little while ago and what I read aligns with your explanation above. Learning new things everyday :)

jestapinski:
No worries! Definitely a bit niche so please feel free to reach out with any questions you might have. The goal here is to properly use the `og:description` tag versus the meta `description` tag.

jestapinski:
> When inspecting articles I see the og metatags there. But the contents are the same as the description metatag so I'm not sure how to best test this. What would be the best way to test the right metadata usage locally?

Oh yeah so let me explain a bit more here. So for the first article, I updated the source rST to have an og:description that SHOULD NOT match the meta description and here we see they do not match:
Screen Shot 2020-08-20 at 3 35 59 PM

The second one, I completely omitted the og description on the source rST and we can see that the og:description tag is still rendered and in this case using the meta description content:
Screen Shot 2020-08-20 at 3 37 55 PM

Apologies because the source rSTs I updated are in the content repo, which is private. I tried to document this as well in the cypress tests, which actually look at the same articles and check these as well.

PR #484, opened 2020-08-25
DEVHUB-202: Add Second RSS Feed for Text Filter

jestapinski:
[JIRA](https://jira.mongodb.org/browse/DEVHUB-202)
[Staging RSS Feed File](https://docs-mongodbcom-staging.corp.mongodb.com/master/devhub/jordanstapinski/DEVHUB-202/search-rss.xml)

Deepak has asked for a second DevHub RSS feed, this time adding some custom fields relating to search. Initially I was thinking of adding to the existing feed but we decided to just keep this separate should we want to change the Search logic (he will be pulling from the generated XML file).

I've verified this format works for Deepak so we should be good on when to include/exclude fields, but will leave some PR comments.

To do so, we did the following things:
- Add the new fields `type`, `authors` (with name), and `tags` for an article into its GraphQL node
- Create a new query to query GraphQL for these additional fields for the new RSS feed
- Add new serializer to take GraphQL query response and format into the XML we want
- Hook serializer and config to `gatsby-config` and `gatsby-config.prod`

Check out the above staging RSS file to see it in action!

### DIFF
--- a/src/utils/setup/serialize-search-rss-data.js
+++ b/src/utils/setup/serialize-search-rss-data.js
@@ -1,17 +1,25 @@
+const getCustomRSSElements = article => {
+    const authorNames = article.authors
+        ? [...article.authors.map(a => ({ author_name: a.name }))]
+        : [];
+    const tags = article.tags ? [...article.tags.map(t => ({ tag: t }))] : [];
+    const customElements = [{ type: article.type }];
+    if (authorNames.length) {
+        customElements.push({ author_names: authorNames });
+    }
+    if (tags.length) {
+        customElements.push({ tags: tags });
+    }
+    return customElements;
+};
+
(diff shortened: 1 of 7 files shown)


src/utils/setup/serialize-search-rss-data.js
```diff
@@ -0,0 +1,25 @@
+const getCustomRSSElements = article => {
```
jestapinski, line 1: Here we want to add some "custom" elements to the feed. We also just want to exclude a tag outright if it has 0 entries (such as no tags for whatever reason). If we don't intentionally exclude, the XML gives an odd effect of just giving `</tagname>` which causes parsing issues.

jestapinski:
Thanks for taking a look!

PR #488, opened 2020-08-26
DEVHUB-204 Part 1: UI Scaffolding

jestapinski:
[InVision](https://mongodb.invisionapp.com/share/5GWV0HJSXZR#/screens/415163294)
[Staging Link (learn page)](https://docs-mongodborg-staging.corp.mongodb.com/master/devhub/jordanstapinski/js/text-filter-layout/learn)

This PR begins some work to implement the new learn page text filter (see InVision). The goal of this project is to add a new filter for the learn page which will hit a backend to fetch articles to show based on a text query (search, but we are not calling it search). More than happy to give additional context on the project! 

This does some small logic rearrangements on the learn page to be able to render content returned from a text filter query.

Before, we rendered content based on which learn page tab was selected. I moved that logic and replaced this with a `switch` statement, since that's effectively what we were doing before.

I also added a new state variable which we will use to get/set the results of a text filter called `textFilterResults`. Yes, we will be showing *only* the results of the text filter query, we will ignore the other filters altogether.

I also added a basic empty-state (to be filled in) so we can see how this will all tie together. The next step will be adding logic to fetch the results of a text filter query and display.

This PR was built to be able to be merged/deployed without any visible changes (text filter is not yet active), so:

### To Test
- Check that the staging link learn page behaves identically to the existing learn page

### DIFF
--- a/src/components/dev-hub/empty-text-filter-results.js
+++ b/src/components/dev-hub/empty-text-filter-results.js
@@ -1,0 +1,6 @@
+import React from 'react';
+
+// TODO: replace with real empty state
+const EmptyTextFilterResults = () => <p>No results</p>;
+
+export default EmptyTextFilterResults;
--- a/src/components/dev-hub/filter-bar.js
+++ b/src/components/dev-hub/filter-bar.js
@@ -57,7 +57,14 @@ const SelectWrapper = styled('div')`
 `;
 
 export default React.memo(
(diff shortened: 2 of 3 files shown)


jestapinski:
I noticed in the staging learn page the videos are rendering then disappearing. This is a separate issue I am going to look into, relating to how we query this content from Realm. My guess is we need to add a specific domain.

jestapinski:
Thanks!

PR #490, opened 2020-08-26
add useFetch debounce hook

jestapinski:
This PR adds a new hook into the library: `useFetch`

This hook is similar in purpose to the one we have for university but adds the ability to debounce the fetch if requested by passing an optional third param.

For tests, I added the `react-hooks-testing-library` and started with the tests we have on University. I then added a debounce test as well.

We are not using this hook anywhere yet.

### DIFF
--- a/src/hooks/use-fetch.js
+++ b/src/hooks/use-fetch.js
@@ -7,6 +7,12 @@ const getErrorMessage = (response, method) =>
         ? `${method} Error: ${response.status} - ${response.statusText}`
         : `${method} Error: Did not receive a response from the server`;
 
+/**
+ * Hook with option to debounce to fetch data (GET) and provide any postprocessing
+ * @param {*} url The url to fetch from
+ * @param {*} postprocessData Manipulations to data to be done after a fetch
+ * @param {*} debounceTime (Optional) if provided, will debounce the fetch based on this number of ms
+ */
 function useFetch(url, postprocessData, debounceTime) {
     const [data, setData] = useState(null);
     const [error, setError] = useState(null);
--- a/src/hooks/use-fetch.js
+++ b/src/hooks/use-fetch.js
@@ -1,0 +1,53 @@
+import { useCallback, useEffect, useState } from 'react';
+
+const isBadResponse = response => !response || (response && !response.ok);
+
+const getErrorMessage = (response, method) =>
+    response
+        ? `${method} Error: ${response.status} - ${response.statusText}`
+        : `${method} Error: Did not receive a response from the server`;
+
+function useFetch(url, postprocessData, debounceTime) {
+    const [data, setData] = useState(null);
+    const [error, setError] = useState(null);
+    const [fetchEvent, setFetchEvent] = useState(null);
+    const fetchData = useCallback(async () => {
+        const resp = await fetch(url);
+        if (isBadResponse(resp)) {
+            console.log('ERROR FOUND');
+            throw new Error(getErrorMessage(resp, 'GET'));
+        }
+        const parsedData = await postprocessData(resp);
+        if (parsedData) {
+            setError(null);
+            setData(parsedData);
+        }
+    }, [postprocessData, url]);
+    useEffect(() => {
+        const getData = async () => {
+            if (debounceTime) {
+                if (fetchEvent) {
+                    clearTimeout(fetchEvent);
+                }
+                setFetchEvent(setTimeout(fetchData, debounceTime));
+            } else {
+                try {
+                    fetchData();
+                } catch (e) {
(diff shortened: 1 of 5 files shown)


tests/hooks/use-fetch.test.js
```diff
@@ -0,0 +1,83 @@
...
+        const { result, rerender, waitForNextUpdate } = renderHook(() =>
+            useFetch(url, x => x.json(), 1000)
+        );
+        expect(window.fetch).toHaveBeenCalledTimes(0);
+        let { data, error } = result.current;
+        expect(data).toEqual(null);
+        expect(error).toEqual(null);
+        // Update the hook 10 additional times before running timers,
+        // only one call should occur (last one)
+        for (let i = 0; i < 10; i++) {
```
jbouzi12, line 70: Interesting, how does this help with testing the debounce functionality?

src/hooks/use-fetch.js
```diff
@@ -0,0 +1,55 @@
...
+    const [fetchEvent, setFetchEvent] = useState(null);
+    const fetchData = async () => {
+        const resp = await fetch(url);
+        if (isBadResponse(resp)) {
+            const errorMessage = getErrorMessage(resp, 'GET');
+            console.warn(errorMessage);
+            setError(errorMessage);
+            setData(null);
+        } else {
+            const parsedData = await postprocessData(resp);
```
jbouzi12, line 28: Curious about the approach here to make `postprocessData` an async function. Is this meant to do things like parse the response as json?

tests/hooks/use-fetch.test.js
jestapinski: Ah so what this does is the `rerender` function will call a new `useEffect` each time (since the url has been updated in the loop). That `useEffect` then sets the timeout to actually fetch the data. With jest we can `useFakeTimers` so this batches the events up so we don't have timers going off until we want. Together (if I did this right) we can keep calling the internal `useEffect` which keeps setting (and clearing) timeouts and so when we run the timers we only hit `fetch` once with the last url.

src/hooks/use-fetch.js
jestapinski: Yep that is my thinking, would it make sense to just maybe return `data` in this case?

tests/hooks/use-fetch.test.js
jbouzi12: That definitely makes sense, really like this approach 👍

src/hooks/use-fetch.js
jbouzi12: If we are just using this to parse the response as json I wonder if we can ensure that the api's return json so this step can be skipped. That way we won't need to convert the data to json and can just return it directly.

But if there are some other async operations we want to do on the data before returning it then using this `postProcessData` approach definitely makes sense. What do you think?

jestapinski: Ah I think that is a great idea, lets do the `json` here like we do in University and postprocess via another hook when this is to be used. I went ahead and made the change

src/hooks/use-fetch.js
```diff
@@ -25,10 +24,10 @@ function useFetch(url, postprocessData, debounceTime) {
             setError(errorMessage);
             setData(null);
         } else {
-            const parsedData = await postprocessData(resp);
-            if (parsedData) {
+            const jsonData = await resp.json();
```
jbouzi12, line 27: I wonder if se can skip this step altogether. Do you know if the api's we'll be using already return json?

jestapinski: Ah [here](https://webhooks.mongodb-realm.com/api/client/v2.0/app/devhubsearch-hgphu/service/devhub_search/incoming_webhook/devhub_search?q=java) is the endpoint we will be using (at least next for the text filter project). It does give us json but maybe a more appropriate idea is to just return the response in this hook and also use `.json()` in another custom hook specifically for the text filter data manipulation. What do you think about that?

jestapinski:
I actually think for now I am going to run this logic instead through our stitch app to handle the postprocessing, etc. so will reopen should the use case for this arise. Thanks for taking a look Jensen!

src/hooks/use-fetch.js
jbouzi12: Ah I see, yea that makes sense. If you find more cases of us needing to convert data to json down the line we can always put `.json()` back in this hook at that point.

jbouzi12:
> I actually think for now I am going to run this logic instead through our stitch app to handle the postprocessing, etc. so will reopen should the use case for this arise. Thanks for taking a look Jensen!

Sounds good!

PR #491, opened 2020-09-01
Add useTextFilter Hook for Fetching via Realm App

jestapinski:
This PR adds a new custom hook called `useTextFilter`. The goal of this custom hook is to use a given text filter query and fetch the response articles.

I added the logic for getting the filtered articles into our DevHub Realm (Stitch) App as a new function called `fetchTextFilterResults`. This function takes one argument (query) and hits the external API to fetch results with this query and parse as JSON before returning. To see it, check out `functions` in the Stitch app! It is pretty straightforward.

I also added some debounce logic and the ability to test by mocking the stitch browser sdk package. I also brought in the `react-hooks-testing-library` package as well.


tests/hooks/utils/mock-text-filter-fetch.js
```diff
@@ -0,0 +1,64 @@
+export const JAVA_RESULTS = [
```
jestapinski, line 1: This module mocks API responses for text filter results based on query (or call to callFunction with specific args)

jestapinski:
Thanks Jensen!

PR #494, opened 2020-09-01
DEVHUB-202: Add Four Missing Text Filter RSS Fields

jestapinski:
For [DEVHUB-202](https://jira.mongodb.org/browse/DEVHUB-202)
[Staging RSS Feed link](https://docs-mongodbcom-staging.corp.mongodb.com/master/devhub/jordanstapinski/js/rss-updates/search-rss.xml)
[Master RSS link](https://docs-mongodbcom-staging.corp.mongodb.com/master/devhub/jordanstapinski/staging/search-rss.xml)

This PR adds four fields we also need for the text filter work into the RSS feed:
- `atf-image`: the location of the image shown with each article
- `slug`: the location of the article
- `products`: the list of products tags
- `languages`: the list of languages tags

To Verify:
- Check out the staging RSS feed link and compare with master RSS feed link (pick a specific article)
- The only changes should be the additions of these four fields. Nothing else should be different.

Here I looked at one for a COVID article to help verify. We can see the only changes were new field additions, as expected.
Screen Shot 2020-09-01 at 3 29 32 PM

### DIFF
--- a/src/queries/search-article-rss-data.js
+++ b/src/queries/search-article-rss-data.js
@@ -5,8 +5,11 @@ const searchArticleRSSData = `
               authors {
                 name
               }
+                atfimage
                 slug: id
                 description
+                languages
+                products
                 pubdate
                 tags
                 title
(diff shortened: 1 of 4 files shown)


jestapinski:
Thanks Tahiya!

PR #498, opened 2020-09-02
Disable Dropdown Selects when Text Filtering

jestapinski:
[Staging Link (Learn Page)](https://docs-mongodborg-staging.corp.mongodb.com/master/devhub/jordanstapinski/js/disabled-selects/)

This PR adds a small but nice piece of functionality in conjunction with the new text filter. When the user is using the text filter, the other filter types (product, language) are unavailable since we are filtering from an external service. This PR disables those filters when unavailable to provide a better experience for the user.

Screen Shot 2020-09-02 at 3 50 17 PM

To verify:
- When the text filter input is empty, the dropdowns next to it should be enabled
- When the text filter input is not empty, the dropdowns next to it should be disabled

### DIFF
--- a/src/components/dev-hub/filter-bar.js
+++ b/src/components/dev-hub/filter-bar.js
@@ -145,6 +145,7 @@ export default React.memo(
                     <FilterLabel>Filter By</FilterLabel>
                     <SelectWrapper>
                         <Select
+                            enabled={!textFilterQuery}
                             narrow
                             name="product"
                             choices={products}
@@ -156,6 +157,7 @@ export default React.memo(
                     </SelectWrapper>
                     <SelectWrapper>
                         <Select
+                            enabled={!textFilterQuery}
                             narrow
                             name="language"
                             choices={languages}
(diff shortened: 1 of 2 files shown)


jestapinski:
> When you already selected from the dropdown and then search, should we have the dropdowns go back to a default state?

Good question, when the user clears their search the filters will still be applied so we could either temporarily reset them or just leave them and re-enable them when the search empties. I can also consult Dan on this but do you also have any thoughts?

jestapinski:
I just spoke to Dan and we are also going to reset the filters in this case, so will add that work in here shortly!

jestapinski:
> I just spoke to Dan and we are also going to reset the filters in this case, so will add that work in here shortly!

Just kidding, Deepak is good with this approach since the filter here is still applied when a user exists the text filter search. This PR is still good for a look as-is

melissarudofsky (approved):

jestapinski:
Thanks! Will merge into staging once the base branch is in

jestapinski:
Actually going to merge this into the feature branch to run the tests

PR #499, opened 2020-09-03
Learn Page Integration Testing (Part 1)

jestapinski:
This PR adds some integration testing for the Learn page.

Previously, we did not have any tests for this page, so now that the text filter work is nearly done and the media component has been built, now seems like a good time to add Cypress in.

One of the challenges for the learn page is to mock Stitch responses for our various data sources (such as videos, podcasts, text filter responses). Cypress lets us stub an individual route using `server` and `route`, but with Stitch the distinction between what function we want to call is in the request body. Cypress doesn't let us case on the request body using `route` (as far as I have found), but they just released a [`route2`](https://docs.cypress.io/api/commands/route2.html#Comparison-to-cy-route) which has this functionality. I tried `route2` for a bit but [had some issues stubbing external requests.](https://github.com/cypress-io/cypress/issues/8487)

For now, I figured testing the text filter work is most important, so I chose to mock the text filter response and will come back for a part 2 with the other functions mocked as well (hopefully using `route2` when that works as expected).

### To Test
Once this PR is against staging the tests will automatically run, but if you want to take a look before:
- Set `.env.production` to use `jordanstapinski` as the `GATSBY_PARSER_CI_USER`
- Run `npm run buildTest`
- Run `npm run serveTest`
- Run the cypress tests


tahiya-mongo (approved):
lgtm!

cypress/support/commands.js
```diff
@@ -69,6 +92,22 @@ Cypress.Commands.add('mockEventsApi', () => {
     cy.route('**/api/event?status=Live', '@liveEventData').as('getLiveEvents');
 });
 
+Cypress.Commands.add('mockTextFilterResponse', () => {
+    cy.fixture('javaTextFilterResponse.json').as('javaTextFilterResponse');
+    cy.server();
+    cy.route({
```
tahiya-mongo, line 98: Do you mind explaining what's going on here? Just to confirm my understanding and expand my knowledge :)

jestapinski: Absolutely, so for the new text filter on the learn page, what happens is when a user enters a search term we send that search term out to our Realm app which then responds with a set of matching articles for that text filter query.

For testing, we don't exactly want to keep hitting Realm because:
1. Results may change which would break testing
2. Realm may go down which would cause tests to fail

Here, what we do is `stub` the response to this specific endpoint (which corresponds with calling the Realm function for the text filter) with some sample data in `javaTextFilterResponse.json` to populate the articles, letting us not need to rely on hitting the production Realm App every time for results!

Does this make sense?

jestapinski:
Thanks!

cypress/support/commands.js
tahiya-mongo: Yep! Thanks!

PR #501, opened 2020-09-03
DEVHUB-182: Keyboard-Accessible Article TOC

jestapinski:
[Staging Article Link](https://docs-mongodbcom-staging.corp.mongodb.com/master/devhub/jordanstapinski/js/toc-keyboard/article/johns-hopkins-university-covid-19-data-atlas)
[^ Article on Prod](https://developer.mongodb.com/article/johns-hopkins-university-covid-19-data-atlas)
[JIRA Ticket](https://jira.mongodb.org/browse/DEVHUB-182)

This PR adds a prop to the tooltip from `react-tiny-popover` which allows the popup menu content to be inserted into the DOM at an appropriate location, allowing the content to be tabbable at a logical point. Before, the popover content was placed at the end of `<body>`, so it would be the last elements a user could tab. Now it is placed right where the `Trigger` is, making for a more smoother experience.

One issue I would like some feedback on is addressing some of the related CSS. I wanted to use a `span` for keeping track of the content ref, but this caused some issues with other elements which rely on the `span` implementation of popover (so now some `span` css was being applied twice). I wasn't sure of the best way to solve this so any thoughts there would be great!

As far as I remember, the tooltip is only used on the article page for the TOC, Share menu, and bottom share link. Anywhere else to check?


src/components/dev-hub/article-share-footer.js
```diff
@@ -28,6 +28,10 @@ const BlogShareLinks = styled('div')`
     a:not(:last-of-type) {
         margin-right: ${size.medium};
     }
+    /* For the tooltip, there is a wrapper span to give a proper tab UX */
+    span > span:first-child {
+        margin-right: 0;
```
jestapinski, line 33: This is a bit wonky, but we want to make sure the margin applied on line 29 for the footer is not double-applied

melissarudofsky: do you know why we have span on line 27? just inspecting the page I don't really see any reason for it. do we need that? if so, can we just apply those styles directly?

melissarudofsky (approved):
looks good!

src/components/dev-hub/article-share-footer.js
jestapinski: Good point! The copy link button is actually a span element so we need to give some space between that and the next link. I actually realized because it is set up this way, we can just simplify the logic to only target `a` tags, which will only target the last three links and give left margin instead

jestapinski:
Thank you!

PR #504, opened 2020-09-10
Add aria-label tag for selects

jestapinski:
[Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/devhub/jordanstapinski/js/aria-label-dropdowns/)

I was running an accessibility audit on the learn page and found we should provide better labeling on the dropdowns for screen-reading purposes. This short PR adds the `aria-label` attribute to the `Select` component to do so by default from now on.


tahiya-mongo (approved):
lgtm!

jestapinski:
🤙

PR #505, opened 2020-09-10
DEVHUB-192: Add Events/Live Events to Realm

jestapinski:
WIP until we can stub Realm properly in Cypress


jestapinski:
Will reopen when Cypress can stub Stitch properly

PR #506, opened 2020-09-14
DEVHUB-175: Automatically Add `tck` Param to Atlas Signup Urls

jestapinski:
[JIRA](https://jira.mongodb.org/browse/DEVHUB-175)
[Staging Link (Relevant article)](https://docs-mongodbcom-staging.corp.mongodb.com/master/devhub/jordanstapinski/js/cloud-attributions/how-to/search-engine-using-atlas-full-text-search)
[Other Staging Article](https://docs-mongodbcom-staging.corp.mongodb.com/master/devhub/jordanstapinski/js/cloud-attributions/article/build-newsletter-website-mongodb-data-platform)

This PR adds some build-time logic to add an attribution `tck` query param to a specific URL supplied in the ticket `https://www.mongodb.com/cloud/atlas/signup`. There are a couple interesting cases that can come up:
- The url might already have some query params on it, so we want to be sure to preserve those
- The url might not have query params yet, so we want to be sure to add an `?` where needed
- We want to be sure to `encodeURIComponents`

~The actual `tck` value might change, but the logic is where we want it to be.~
The `tck` value is the last piece of the URL slug (aka the article title (but not always the title, we just use whatever the end of the slug is))

To Verify:
- Check out the two article staging links
- If you open the inspector console and cmd+f for signup, you should see the `tck` query param
- If you click these links the tck should stay, as well as any other query params


src/utils/setup/update-attribution-links.js
```diff
@@ -0,0 +1,31 @@
...
+    // Search through ast.children
+    const children = dlv(doc, 'ast.children', []);
+    const findAndUpdateAtlasAttributionRefs = node => {
+        if (node['type'] === 'reference') {
+            if (node.refuri && node.refuri.includes(ATLAS_SIGNUP_LINK)) {
+                let oldSignupBaseUrl = ATLAS_SIGNUP_LINK;
+                let suffix = `tck=devhub_${atlasTckValue}`;
+                // If there are already query params, we want to be sure to
+                // preserve them, including the leading "?"
+                if (node.refuri.includes(`${ATLAS_SIGNUP_LINK}?`)) {
```
jestapinski, line 14: I can probably make this logic a bit cleaner, any ideas?

tahiya-mongo: Didn't test this, but would something like this work?
```
const encoded_suffix = node.refuri.includes(`${ATLAS_SIGNUP_LINK}?`) ? `${encodeURIComponent(`${suffix}&`)}` : `?${encodeURIComponent(`${suffix}&`)}`;
node.refuri = ATLAS_SIGNUP_LINK + encoded_suffix;
```

tahiya-mongo:
Looked through the staged articles and things look good to me.

src/utils/setup/update-attribution-links.js
jestapinski: Ah I tried this out for a few minutes, but we run into one issue.

So some articles have these urls embedded already with query params, such as one like this `https://www.google.com/url?q=https://www.mongodb.com/cloud/atlas/signup?form-type%3Dfree&sa=D&ust=1579301554143000&usg=AFQjCNHI71SacwhKWJLcHMUa9RvY5bAoKw` so we also want to preserve those. This solution is close but gives us this:

`https://www.google.com/url?q=https://www.mongodb.com/cloud/atlas/signup?tck%3Ddevhub-search-engine-using-atlas-full-text-search%26?form-type%3Dfree&sa=D&ust=1579301554143000&usg=AFQjCNHI71SacwhKWJLcHMUa9RvY5bAoKw` because when we `replace` we are missing the extra `?` to replace, so we have an extra `?` in the query params.

I think the way we have it here is perhaps the best we can do for now given this unless there is another thought?

tahiya-mongo: Ah! I see. That makes sense. Yeah I guess this is the best way.

jestapinski:
I think this is good for another look, in the ticket we decided on having the `tck` be `devhub-<filename>` so I updated for that and added a quick cypress test

tahiya-mongo (approved):
Took another look. lgtm! Feel free to wait for either Melissa or Jensen to take a quick look.

jestapinski:
> Took another look. lgtm! Feel free to wait for either Melissa or Jensen to take a quick look.

Thanks! I'll let this sit for a day-ish if anyone wants to take another look and likely merge tomorrow afternoon. Thanks again!

jestapinski:
Thanks Tahiya!

PR #510, opened 2020-09-15
Bump card list default to 12

jestapinski:


jestapinski:
Will be adding in a separate PR

PR #516, opened 2020-09-24
DEVHUB-222 & DEVHUB-184: "Load More" with Query Params

jestapinski:
[JIRA 222](https://jira.mongodb.org/browse/DEVHUB-222)
[JIRA 184](https://jira.mongodb.org/browse/DEVHUB-184)
[Staging Link](https://docs-mongodborg-staging.corp.mongodb.com/master/devhub/jordanstapinski/DEVHUB-222/learn)

This PR does a few things related to the load more button on card lists to make it more SEO friendly. Specifically:
1. The Load More button is now a link which appends a query param to the end of the current URL which will control how many items are shown
2. Implements the `shouldUpdateScroll` Gatsby Browser API method to keep the scroll position constant if a link goes to the same page (such as adds a query param)

A/C:
- A new `page` query param pops up when clicking "load more"
- Scroll position does not change
- The other query params still exist and work around `page` for filtering
- Also applies to "Tag" pages

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "@leafygreen-ui/checkbox": "^4.1.1",
     "@leafygreen-ui/code": "^3.5.0",
     "@leafygreen-ui/icon": "^6.4.2",
+    "@reach/router": "^1.3.4",
     "@testing-library/react-hooks": "^3.4.1",
     "copy-to-clipboard": "^3.3.1",
     "css-loader": "^4.3.0",
--- a/src/components/dev-hub/card-list.js
+++ b/src/components/dev-hub/card-list.js
@@ -102,7 +102,6 @@ const renderContentTypeCard = (item, openAudio) => {
 export default React.memo(({ videos, articles, podcasts, limit = 9 }) => {
     const location = useLocation();
     const { pathname, search } = location;
-    console.log(location);
     // Get page if exists from search
     // Build next link, preserving other links
     const nextPageLink = useMemo(() => {
--- a/src/components/dev-hub/card-list.js
+++ b/src/components/dev-hub/card-list.js
@@ -11,6 +11,8 @@ import { getNestedText } from '../../utils/get-nested-text';
 import { getTagLinksFromMeta } from '../../utils/get-tag-links-from-meta';
 import getTwitchThumbnail from '../../utils/get-twitch-thumbnail';
 
+const CARD_LIST_LIMIT = 12;
+
 const CardContainer = styled('div')`
     display: grid;
     grid-template-columns: repeat(auto-fill, 350px);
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -34,6 +34,7 @@
     "@leafygreen-ui/checkbox": "^4.1.1",
     "@leafygreen-ui/code": "^3.5.0",
     "@leafygreen-ui/icon": "^6.4.2",
+    "@reach/router": "^1.3.4",
```
melissarudofsky, line 37: Curious about your decision to add `"@reach/router"` here instead of using your current solution for routing. Was this something you couldn't do with your current implementation?

src/components/dev-hub/card-list.js
```diff
@@ -97,49 +101,72 @@ const renderContentTypeCard = (item, openAudio) => {
...
-            )}
-            {podcasts.length ? (
-                <Audio onClose={closeAudio} podcast={activePodcast} />
-            ) : null}
-        </>
-    );
-});
+export default React.memo(
+    ({ videos, articles, podcasts, limit = CARD_LIST_LIMIT }) => {
+        const location = useLocation();
```
melissarudofsky, line 106: Can simplify to `const { pathname, search } = useLocation()` ?

src/components/dev-hub/card-list.js
```diff
@@ -97,49 +101,72 @@ const renderContentTypeCard = (item, openAudio) => {
...
+        const { pathname, search } = location;
+        const localPage = pathname.replace(__PATH_PREFIX__, '');
+        // Get page if exists from search
+        // Build next link, preserving other links
+        const nextPageLink = useMemo(() => {
+            const { page, ...params } = parseQueryString(search);
+            if (page) {
```
melissarudofsky, line 113: This was slight hard to understand. Can we simplify the logic by doing something like
`
const pageNumber = page ? parseInt(page) + 1 : 2;
return localPage + buildQueryString({ page: PageNumber, ...params });
`

src/pages/learn.js
```diff
@@ -103,6 +103,9 @@ const filterArticles = (filter, initialArticles) => {
     const filterValues = Object.keys(filter);
     return initialArticles.reduce((acc, article) => {
         for (let i = 0; i < filterValues.length; i++) {
+            if (filterValues[i] === 'page') {
```
melissarudofsky, line 106: cool!

src/pages/learn.js
```diff
@@ -207,6 +210,25 @@ export default ({
         filter => filterArticles(filter, initialArticles),
         [initialArticles]
     );
+    // Update the filter value for page so it behaves nicely with query params
+    const updatePageFilter = useCallback(
+        search => {
+            const { page } = parseQueryString(search);
+            if (page) {
+                filterValue['page'] = page;
+                setFilterValue({ ...filterValue });
```
melissarudofsky, line 219: spread necessary here?

melissarudofsky (commented):
everything looks like it works well! a few small questions!

src/components/dev-hub/card-list.js
jestapinski: Ah yeah I think I had done this for debugging to look at everything in `location`

package.json
jestapinski: So [`reach/router` is already a co-dependency of Gatsby](https://www.gatsbyjs.com/docs/reach-router-and-gatsby/) meaning we already had it installed as a dependency but are now just indicating it is not a dev one (lockfile change is super small too). Gatsby gives a `location` prop to each [page component,](https://github.com/mongodb/devhub/pull/520) so the learn page component has a `location` prop provided but we also want the `CardList` component to have this as well so it can add the query param.

I wanted to go with this approach of adding the hook instead of passing the location down since it made the props a bit cleaner and then we wouldn't have to go back and update every `CardList` with a location.

So `reach/router` just makes things a bit cleaner instead of prop drilling but in theory we don't need it.

src/pages/learn.js
jestapinski: Yeah I think we keep adding filters so maybe we will revisit this piece of functionality soon to clean things up for all filter types

package.json
melissarudofsky: makes sense! thank you for the explanation

jestapinski:
Alright I went ahead and cleaned up the `CardList` as well as updated the above staging link so should be good for another look!

melissarudofsky (approved):

jestapinski:
Thanks Melissa!

PR #518, opened 2020-09-25
DEVHUB-225: Use a Query Param for Text Filter

jestapinski:
[JIRA](https://jira.mongodb.org/browse/DEVHUB-225)
[Staging Link](https://docs-mongodbcom-staging.corp.mongodb.com/master/devhub/jordanstapinski/DEVHUB-225/learn/)

This PR adds a small change to use a query param to keep track of the TextFilter value. Previously we were not doing so, so by keeping track of it with the rest of the existing learn page filters we can add it to the query params basically for free.

I also added a cypress test to model this behavior

A/C:
- There should be a `text` queryParam for the learn page if and only if there is some value in the text filter.
- This query param should be the same as the value in the text filter
- On loading the page with this query param, the filter should be properly updated


tahiya-mongo (approved):
lgtm!

jestapinski:
Thanks!

PR #523, opened 2020-09-30
Fix updated date field

jestapinski:
[Staging Link (Fixed Article)](https://docs-mongodbcom-staging.corp.mongodb.com/master/devhub/jordanstapinski/js/fix-updated-date/article/coronavirus-map-live-data-tracker-charts)
[Prod Link (Broken Article)](https://developer.mongodb.com/article/coronavirus-map-live-data-tracker-charts)

Somewhere along the way we were expecting the name of this field to be `updatedDate` but it was coming in as `updated-date`. This PR swaps to the correct field name and adds a cypress test for it as well.

Note in the broken article there is no updated date in the heading but in the staging link there is.


tahiya-mongo (approved):
lgtm!

jestapinski:
Thanks!

PR #525, opened 2020-10-05
Throw Error if non devhub-article template is given

jestapinski:
[Staging Link](https://docs-mongodborg-staging.corp.mongodb.com/master/devhub/jordanstapinski/js/robust-template/)

re: issue in slack with a malformed template response. This PR removes the unused `document` template and throws an error at build-time if this field is malformed.


src/utils/get-template.js
```diff
@@ -4,7 +4,9 @@ const getTemplate = template => {
         case 'devhub-article':
             return 'article';
         default:
-            return 'document';
+            throw new Error(
+                `The given article template was not devhub-article, it was ${template}. This means there was likely a parsing issue.`
+            );
```
sophstad, line 9: To be crystal clear: throwing this error means that the Gatsby build stops?

jestapinski: Yes, the build is halted

sophstad (approved):

jestapinski:
Thanks!

PR #528, opened 2020-10-08
Tab Directive Support

jestapinski:
[Staging Article Using New Directive](https://docs-mongodbcom-staging.corp.mongodb.com/master/devhub/jordanstapinski/js/tabs/article/sample-tabs-article)
[JIRA](https://jira.mongodb.org/browse/DEVHUB-238)

This PR adds a component to support a new Article directive called `.. tabs::`. [Here is the documentation](https://docs.mongodb.com/meta/tutorials/tabbed-content/) on this directive, but in a nutshell the premise is this:

- An article can use `.. tabs::` and hide content within
- The typical use case is for toggling which programming language to display on code
- **When one tab on a page changes, the others change to match (so programming language stays constant, etc)** (try this one out!)
- Tabs support a `:hidden:` option which hides the tab bar from view (so content toggles)
- Each `tab` has a `tabid` which corresponds to its name. These are mapped in the component to something more pleasant.

Much of this is copied from the Docs Platform front-end and I will show where this is linked. Our implementation of the tabs uses the LeafyGreen component whereas theirs is custom.

I will add Cypress tests for this after we get it in because Deepak wants the writers to try it out in a "beta" fashion since this ask is high priority and we don't support `.. tabs::` at all right now.

Here is the raw rST for the tabs in this article:

```
   .. tabs::

      .. tab::
        :tabid: shell
        .. code-block:: bash
          mongoexport --collection='statistics' --out='statistics.jsonl' --uri="mongodb+srv://readonly:readonly@covid-19.hip2i.mongodb.net/covid19"
          mongoexport --collection='metadata' --out='metadata.jsonl' --uri="mongodb+srv://readonly:readonly@covid-19.hip2i.mongodb.net/covid19"

      .. tab::
        :tabid: cpp
        .. code-block:: cpp
          Some C++ code

      .. tab::
        :tabid: php
        .. code-block:: php
          Some PHP code

   Now for some additional tabs

   .. tabs::
      :hidden:

      .. tab::
        :tabid: shell
        The reader selected bash

      .. tab::
        :tabid: cpp
        The reader selected CPP

      .. tab::
        :tabid: php
        The reader selected PHP

```

### DIFF
--- a/cypress/integration/tag.js
+++ b/cypress/integration/tag.js
@@ -18,7 +18,7 @@ describe('Tag page', () => {
         });
     });
     it('should contain several articles with basic information', () => {
-        cy.get('[data-test="card"]').should('have.length', 4);
+        cy.get('[data-test="card"]').should('have.length', 5);
         cy.get('[data-test="card"]')
             .first()
             .then(card => {
--- a/src/components/dev-hub/button.js
+++ b/src/components/dev-hub/button.js
@@ -154,9 +154,20 @@ const StyledButton = styled('button')`
 
     ${({ primary, theme }) => primary && primaryStyles(theme)}
     ${({ secondary, theme }) => secondary && secondaryStyles(theme)}
-    ${({ play, theme }) => play && playStyles(theme)}
-    ${({ play, primary, secondary, theme }) =>
-        !primary && !secondary && !play && tertiaryStyles(theme)}
+    ${({
+        play,
+        theme,
+    }) => play && playStyles(theme)}
+    ${({
+        play,
+        primary,
+        secondary,
+        theme,
+    }) =>
+        !primary &&
+        !secondary &&
+        !play &&
+        tertiaryStyles(theme)}
     &[disabled],
     &[disabled]:hover {
         background: ${({ theme }) => theme.colorMap.greyLightThree};
@@ -187,7 +198,12 @@ const getArrow = ({ pagination, primary, secondary }) => {
  */
 
 const Button = ({ children, href, play, to, hasArrow = true, ...props }) => {
-    const isButton = !!(props.primary || props.secondary || play);
+    const isButton = !!(
+        props.primary ||
+        props.secondary ||
+        props.renderAsButton ||
+        play
+    );
     const arrow = hasArrow ? getArrow(props) : null;
     if (href || to || !isButton) {
         // If the Button has a `to` or a `href` prop, then it renders as a `Link` element,
--- a/src/components/dev-hub/tab-context.js
+++ b/src/components/dev-hub/tab-context.js
@@ -1,0 +1,6 @@
+import React from 'react';
+
+export const TabContext = React.createContext({
+    activeTabs: undefined,
+    setActiveTab: () => {},
+});
--- a/src/components/dev-hub/tab-context.js
+++ b/src/components/dev-hub/tab-context.js
@@ -1,4 +1,4 @@
-import React, { useEffect, useReducer } from 'react';
+import React, { useCallback, useEffect, useReducer, useState } from 'react';
 import { getLocalValue, setLocalValue } from '../../utils/browser-storage';
 
 const defaultContextValue = {
@@ -15,9 +15,25 @@ const reducer = (tabSelectionPreferences, { tabset, value }) => ({
 });
 
 const TabProvider = ({ children }) => {
+    const [debouncedTabUpdate, setDebouncedTabUpdate] = useState(null);
     const [activeTabs, setActiveTab] = useReducer(
         reducer,
         getLocalValue('activeTabs') || defaultContextValue.activeTabs
+    );
+
+    /*
+      The LeafyGreen Tabs Implementation keeps adding key event handlers
+      to the DOM. By using a setTimeout we can avoid duplicate events being
+      run and so key events work as expected
+    */
+    const debouncedSetActiveTab = useCallback(
+        v => {
+            if (debouncedTabUpdate) {
+                clearTimeout(debouncedTabUpdate);
+            }
+            setDebouncedTabUpdate(setTimeout(() => setActiveTab(v)));
+        },
+        [debouncedTabUpdate]
     );
 
     useEffect(() => {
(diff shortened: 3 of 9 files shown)


src/components/dev-hub/tab-context.js
```diff
@@ -0,0 +1,33 @@
+import React, { useEffect, useReducer } from 'react';
```
jestapinski, line 1: Pulled directly from https://github.com/mongodb/snooty/blob/a37358d314fc0468acb5bf82fc4f8919279ddb76/src/components/tab-context.js

src/components/dev-hub/tabs.js
```diff
@@ -0,0 +1,127 @@
+import React, { useCallback, useContext, useEffect, useState } from 'react';
```
jestapinski, line 1: [Docs FE starting Tab code](https://github.com/mongodb/snooty/blob/a37358d314fc0468acb5bf82fc4f8919279ddb76/src/components/Tabs.js)

src/utils/browser-storage.js
```diff
@@ -0,0 +1,30 @@
+import { isBrowser } from './is-browser';
```
jestapinski, line 1: Basically identical to https://github.com/mongodb/snooty/blob/a37358d314fc0468acb5bf82fc4f8919279ddb76/src/utils/browser-storage.js

cypress/integration/tag.js
```diff
@@ -18,7 +18,7 @@ describe('Tag page', () => {
         });
     });
     it('should contain several articles with basic information', () => {
-        cy.get('[data-test="card"]').should('have.length', 4);
+        cy.get('[data-test="card"]').should('have.length', 5);
```
jestapinski, line 21: I added a new article with the rST for `tabs` since this is a new directive into the CI repo. That is why this is one more than before. I will also be using this article for creating cypress tests about the tab directive.

src/components/dev-hub/tab-context.js
```diff
@@ -0,0 +1,33 @@
...
+import { getLocalValue, setLocalValue } from '../../utils/browser-storage';
+
+const defaultContextValue = {
+    activeTabs: {},
+    setActiveTab: () => {},
+};
+
+const TabContext = React.createContext(defaultContextValue);
+
+const reducer = (prevState, { name, value }) => ({
```
melissarudofsky, line 11: Why do we keep track of prevState here?

melissarudofsky (commented):
This is cool! Just wondering, is there anything that checks the RSTs to make sure that all the same tabIds are used in ever tab component? might they ever want to have two or more sets of tabs in one article?

src/components/dev-hub/tab-context.js
jestapinski: Ah that way if we want to add new tabs in we don't overwrite the old one in this object

jestapinski:
> This is cool! Just wondering, is there anything that checks the RSTs to make sure that all the same tabIds are used in ever tab component? might they ever want to have two or more sets of tabs in one article?

Good question! They don't have to use the same ones everywhere (although they should) but there is no specific check that I know of. If they don't perfectly align then they won't get the nice behavior where others update automatically. Maybe we could give some warning but honestly not sure how they want to use it.

We are having a workshop next week with some of the writers on this so we can ask them if they want a more rigorous check on it

src/components/dev-hub/tab-context.js
jestapinski: This piece of state actually wraps the entire app so preferences are kept between articles, and different articles could have different sets of tabs

melissarudofsky: Okay, I see! I totally misunderstood what this was doing. Maybe it would be helpful to rename the name/value variables to better describe what we are actually storing here?

jestapinski: Yep! Great idea, I went with `tabset` to hopefully make that more clear as well as `tabSelectionPreferences`

melissarudofsky (approved):

jestapinski:
Thanks! Found a bug yesterday where the focus always jumps to the last `Tabs` on a change so will fix that up before merging

jestapinski:
I fixed two bugs here and pushed up the changes, I'll point out where in the code changes were made and what happened

src/components/dev-hub/tab-context.js
```diff
@@ -0,0 +1,52 @@
...
+        reducer,
+        getLocalValue('activeTabs') || defaultContextValue.activeTabs
+    );
+
+    /*
+      The LeafyGreen Tabs Implementation keeps adding key event handlers
+      to the DOM. By using a setTimeout we can avoid duplicate events being
+      run and so key events work as expected
+    */
+    const debouncedSetActiveTab = useCallback(
```
jestapinski, line 29: One issue is if we have multiple sets of tabs on the page. Right now LeafyGreen adds a new event handler every time we render a `Tabs` component, meaning if we have three `Tabs` on the page then they add three separate event handlers for a key press.

This led to an issue where pressing an arrow key would not swap tabs properly. The solution for now is to debounce these on the event loop so we only get one event at a time. Should an article use tabs with different sets of tabs from each other, this could not work as anticipated (some information will be lost), but I'm willing to revisit this soon if it is an issue.

[Here](https://github.com/mongodb/leafygreen-ui/blob/master/packages/tabs/src/Tabs.tsx#L167) is the code where they add an event handler once for each tab, causing the problem.

src/components/dev-hub/tabs.js
```diff
@@ -0,0 +1,146 @@
...
+    );
+
+    return (
+        <>
+            <StyledTabs
+                selected={activeTab}
+                setSelected={onClick}
+                darkMode
+                ishidden={hidden}
+                as={TabButton}
```
jestapinski, line 115: Another issue was with this `as` prop. When using the default `button` value, the page would jump on a re-render to the last set of tabs. Sophie and I have no idea what is the root cause, but using anything else here, such as our `Button` does not give this issue so I went ahead and swapped our button in here.

src/components/dev-hub/button.js
```diff
@@ -187,7 +198,12 @@ const getArrow = ({ pagination, primary, secondary }) => {
  */
 
 const Button = ({ children, href, play, to, hasArrow = true, ...props }) => {
-    const isButton = !!(props.primary || props.secondary || play);
+    const isButton = !!(
+        props.primary ||
+        props.secondary ||
+        props.renderAsButton ||
```
jestapinski, line 204: I added this new option for the `Button` so we can render anything as a button without needing to apply `primary` or `secondary` and then removing styles manually.

jestapinski:
Thanks all for your help on this one!

PR #531, opened 2020-10-14
Clean up cypress concurrency issue

jestapinski:
This PR fixes an issue with cypress on the learn page where we were not effectively re-querying the DOM after a re-render ([similar issue to this mitx PR](https://github.com/10gen/mitx/pull/5123) with a similar resolution).

By cleaning this up, we don't need to optionally retry anymore for the following test, and I renamed some of the `describes` for clarity.


jbouzi12 (approved):
This lgtm. The tests themselves are much clearer as well.

jestapinski:
Thanks!

PR #533, opened 2020-10-15
Cypress Tests for Tabs Component

jestapinski:
[Staging Article (the one used in the cypress tests)](https://docs-mongodbcom-staging.corp.mongodb.com/master/devhub/jordanstapinski/js/cypress-tabs/article/sample-tabs-article)

This PR continues with adding tests for the Tabs component by adding Cypress tests for it. Using a sample article, the UI tests check basic rendering and interactions, as well as some of the different options including the `hidden` option. I'd suggest checking out the staging link provided to help see some of these tests in action!

### DIFF
--- a/cypress/integration/article.js
+++ b/cypress/integration/article.js
@@ -36,6 +36,8 @@ const TWITTER_SHARE_URL = `https://twitter.com/intent/tweet?url=${PROD_ARTICLE_U
 const UPDATED_ARTICLE_URL =
     '/article/coronavirus-map-live-data-tracker-charts/';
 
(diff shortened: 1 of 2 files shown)


jbouzi12 (approved):
lgtm!

jestapinski:
That is 100% the intended behavior, the idea is we want the reader to only select their preference one time. In the rST writers can opt-in to this feature by having a field called the `tabid` match up between sets of tabs. This field is separate from the name so in this example even though one has `Bash` and one has `Mongo Shell` they actually have the same `tabid` so this is expected.

The unit tests will highlight this as well.

Thanks for the review!

PR #537, opened 2020-10-26
Chunk asset queries

sophstad:
[[Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/devhub/sophstad/ss/fix-assets/)] Use the aggregation pipeline to split our assets query into chunks of a given size. A subset of results is returned each time, which avoids hitting Realm's memory limit. This approach was significantly faster than running `.find()` multiple times, which proved prohibitively slow.


jestapinski (approved):
Seems a tad slower to me but since this is blocking let's get it out there and address again later

PR #542, opened 2020-11-05
Bump GH Actions plugin versions

jestapinski:
Bumps the following GH Action plugins:
- Cache
- Cypress
- Upload (for artifacts on failure)

The reason this came up was I noticed these deprecation warnings in the CI job generally. [Here is a job with the warnings](https://github.com/mongodb/devhub/actions/runs/345764491).

Screen Shot 2020-11-05 at 1 51 07 PM

We are not using `set-env` in our configs so I assumed this was one of the plugins we use and lo and behold, the CI job on this PR has no such warnings.


jestapinski:
Going to go ahead and merge this in since these are being deprecated on the 16th and want to make sure everything continues to work as expected.

PR #543, opened 2020-11-05
DEVHUB-234: (part 1) Refactor Featured Home Articles Logic

jestapinski:
[Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/devhub/jordanstapinski/js/home-featured-refactor/)
[JIRA](https://jira.mongodb.org/browse/DEVHUB-234)

This PR is related to a larger effort to un-hardcode logic for experiences on the DevHub which only support articles by refactoring Home page "Feature Article" section.

Previously, we had some logic and terminology which suggested this section only supports articles (which it still does right now), but this PR lays the foundation for supporting new content types here by doing the following:

- Adding a new `type` field to featured articles which can be used to distinguish between pieces of content. Ideally these should all share the same API (TS would be awesome for this someday) but until then this can help distinguish between cases
- Adding `switch` statements to better support forthcoming content types here

I also took this chance to remove an old piece of functionality relating to featured content. We used to have `DEFAULT` arrays for home/learn featured content but these haven't been used in a while and could be stale anyway.

### DIFF
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -92,8 +92,7 @@ const DescriptiveText = styled(P)`
 
 // TODO: Generalize as new content types are supported
 const FeaturedHomePageItem = ({ item }) => {
-    const itemType = item.type;
-    if (itemType === 'article') {
+    if (item.type === 'article') {
         const { image, slug, title } = getFeaturedCardFields(item);
         return (
             <StyledTopCard
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -90,7 +90,24 @@ const DescriptiveText = styled(P)`
     margin-bottom: ${size.medium};
 `;
 
-const IndexPageContent = ({ stream, title, twitchVideo, featuredArticles }) => {
+// TODO: Generalize as new content types are supported
+const FeaturedHomePageItem = ({ item }) => {
+    const itemType = item.type;
+    if (itemType === 'article') {
+        const { image, slug, title } = getFeaturedCardFields(item);
+        return (
+            <StyledTopCard
+                maxTitleLines={3}
+                image={image}
+                to={slug}
+                title={title}
+                key={title}
+            />
+        );
+    }
+};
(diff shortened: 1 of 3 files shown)


src/utils/setup/handle-create-page.js
```diff
@@ -12,23 +13,14 @@ let stitchClient;
...
-    'quickstart/java-setup-crud-operations',
-    'how-to/golang-alexa-skills',
-    'how-to/polymorphic-pattern',
-];
-
 const memoizedStagingPages = memoizerific(1)(
     async () => await getStagingPages()
 );
 
+const memoizedBuildTimeMedia = memoizerific(1)(
```
jestapinski, line 20: We are also going to be adding podcasts and videos in the near future so I pre-emptively added this memoization. It isn't used right now so I'm also happy to do this at a later time instead

src/pages/index.js
```diff
@@ -90,7 +90,24 @@ const DescriptiveText = styled(P)`
     margin-bottom: ${size.medium};
 `;
 
-const IndexPageContent = ({ stream, title, twitchVideo, featuredArticles }) => {
+// TODO: Generalize as new content types are supported
+const FeaturedHomePageItem = ({ item }) => {
+    const itemType = item.type;
```
jbouzi12, line 95: Tiny nit, since we are only using `itemType` once here we probably don't need this variable. We can just use `item.type` directly in the conditon.

jbouzi12 (commented):
Just had a small nit, other wise lgtm!

src/pages/index.js
jestapinski: Good catch! I was debating how I wanted to represent the item `type` before settling on an object property (which now makes me think more about adding TS in the future)

tahiya-mongo (approved):
LGTM!

jestapinski:
Thanks all!

PR #544, opened 2020-11-06
DEVHUB-251: Sort CardList by created-date

jestapinski:
[Staging Link (to learn page)](https://docs-mongodbcom-staging.corp.mongodb.com/master/devhub/jordanstapinski/DEVHUB-251/learn/)
[JIRA](https://jira.mongodb.org/browse/DEVHUB-234)

This PR makes a tweak to the general `CardList` component which will now account optionally for an article's `updated-date` when sorting. Most of the work here is improving the end-to-end tests to add a check for this behavior as well as clean up some tests impacted by the ordering change.

### DIFF
--- a/cypress/support/commands.js
+++ b/cypress/support/commands.js
@@ -106,11 +106,11 @@ Cypress.Commands.add('toggleLearnPageTab', tabName => {
     cy.get('[data-test="tabs"]').within(() => {
         cy.contains(tabName).should('exist').click();
     });
-    // Check that this tab is now active before moving on to check content
-    cy.get(`[data-test="tab-${tabName}"]`).should(
-        'have.css',
-        'border-bottom-color'
-    );
+    if (tabName === 'All' || tabName === 'Articles') {
+        cy.get('[data-test="filter-bar"]').should('exist');
+    } else {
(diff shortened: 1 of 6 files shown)


cypress/support/commands.js
```diff
@@ -104,8 +104,12 @@ Cypress.Commands.add('mockTextFilterResponse', () => {
 
 Cypress.Commands.add('toggleLearnPageTab', tabName => {
     cy.get('[data-test="tabs"]').within(() => {
-        cy.contains(tabName).click();
+        cy.contains(tabName).should('exist').click();
     });
+    cy.get(`[data-test="tab-${tabName}"]`).should(
+        'have.css',
+        'border-bottom-color'
+    );
```
jestapinski, line 112: There is probably a better way we can check for if the tab is now active here before moving on (maybe a comment here would do the trick as well?)

tahiya-mongo: Yeah, commenting here should do the trick.

tahiya-mongo (approved):
LGTM!

jestapinski:
The CI test became inconsistent due to re-renders between loading the page and hitting the tab (since we pull in data). I simplified the test greatly which helped remove some lingering issues. Thanks for the look Tahiya!

PR #545, opened 2020-11-09
DEVHUB-252: Support Object Representation for Assets

jestapinski:
[Staging Link](https://docs-mongodborg-staging.corp.mongodb.com/master/devhub/jordanstapinski/DEVHUB-252/)
[JIRA Ticket](https://jira.mongodb.org/browse/DEVHUB-252)

This PR follows the [related Snooty work](https://github.com/mongodb/snooty/pull/315/files) by changing the datatype of an asset from a string to an object containing the following fields:
- `checksum` (essentially a hash of the image)
- `key` (the filepath to store at)

Here is a summary of what I changed:
- Changed our GraphQL representation of an image to also have a `paths` array field (for all of the relevant filepaths to save as)
- Changed our local representation of assets from an Array to an Object mapping the checksum value to filepaths
- Updated testing

### DIFF
--- a/src/utils/setup/save-asset-files.js
+++ b/src/utils/setup/save-asset-files.js
@@ -20,7 +20,7 @@ const saveFile = async (buffer, filepath) => {
 export const saveAssetFiles = async (assets, stitchClient) => {
     const numAssets = Object.keys(assets).length;
     if (numAssets) {
-        const assetQuery = { _id: { $in: Array.from(Object.keys(assets)) } };
+        const assetQuery = { _id: { $in: Object.keys(assets) } };
         const assetCollection = stitchClient
             .getServiceClient(RemoteMongoClient.factory, 'mongodb-atlas')
             .db(DB)
--- a/src/utils/setup/save-asset-files.js
+++ b/src/utils/setup/save-asset-files.js
@@ -9,28 +9,28 @@ const DB = metadata.database;
 
 const CHUNK_SIZE = 500;
 
-const saveFile = async asset => {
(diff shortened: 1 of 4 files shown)


jestapinski:
I expect the CI tests to fail until the CI data has been upgraded to the forthcoming parser version as this is a breaking change

src/utils/setup/save-asset-files.js
```diff
@@ -9,28 +9,25 @@ const DB = metadata.database;
...
 };
 
-// Write all assets to static directory
+// Write all assets to static directory while including all filepaths
 export const saveAssetFiles = async (assets, stitchClient) => {
-    if (assets.length) {
-        const assetQuery = { _id: { $in: assets } };
+    const numAssets = Object.keys(assets).length;
+    if (numAssets) {
+        const assetQuery = { _id: { $in: Array.from(Object.keys(assets)) } };
```
sophstad, line 23: [nit] I believe `Object.keys()` [returns an array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys#Return_value); is there any reason to create a new instance here?

sophstad (approved):
Looks great, just a minor note. Thanks so much for your speedy help! 🚀

src/utils/setup/save-asset-files.js
jestapinski: I ran into some issue with it expecting an array and thought it was coming from here. I think you are right, let me go ahead and double-check that

jestapinski:
Thanks for the help Sophie!

PR #552, opened 2020-11-16
DEVHUB-253: Migrate Community Project Submission

jestapinski:
[Staging Link](https://docs-mongodborg-staging.corp.mongodb.com/master/devhub/jordanstapinski/DEVHUB-253/)
[JIRA](https://jira.mongodb.org/browse/DEVHUB-253)

This PR deals with an under-the-hood change to how we manage the "Submit your project" form on the home page of the DevHub (at the bottom). Previously this was managed by the Docs Platform team Realm app. I did the following to set this up in the DevHub Realm App:

(Not in this PR)
- `mongoexport` existing project data from the Docs Platform Atlas Cluster, and then `mongoimport`ed it into the DevHub Atlas Cluster (there were 12 records)
- Added credentials to the SES service used for emailing `community@mongodb.com` to the DevHub Realm App and set up two functions: a callback to be used on form submission and a trigger to send `community@mongodb.com` an email when an insert is done to the collection we are storing community projects in
- Deployed these non-user-facing-changes via Realm

(This PR)
- Switched the front-end to now point at the DevHub Realm App for submitting these projects
- Removed some old Docs Platform Realm code we are now no longer using

(Can be done after this PR)
- Realm functions relating to the project submission, and Atlas data, can safely be removed from the Docs Platform Realm/Atlas projects

I went through this new workflow twice with Jamie, so I am feeling good about this migration. Testing this can be done by filling out the form at the bottom but this will also store a record in the DB and ping Jamie.


src/utils/devhub-api-stitch.js
```diff
@@ -50,3 +50,13 @@ export const submitAcademiaForm = async academiaData => {
     );
     return result;
 };
+
+export const submitDevhubProject = async (metadata, object, segmentData) => {
+    const result = await callDevhubAPIStitchFunction(
```
tahiya-mongo, line 55: Is there a reason for storing the result in a variable here? Just curious

jestapinski: No reason other than to match the existing structure of the other functions in this file. Previously it just returned the promise directly

tahiya-mongo (approved):
Didn't fill out the form to test, but the changes look good to me!

jestapinski:
Thanks for taking a look @tahiya-mongo ! Yeah we don't necessarily want to spam Jamie but I tested it out with her just to make sure she was getting the notices she expects. This works very similar in practice to what we did for the Academia form back in the day.

PR #555, opened 2020-11-23
DEVHUB-185 and DEVHUB-235: Learn Page Filter State

jestapinski:
[Staging Link](https://docs-mongodborg-staging.corp.mongodb.com/master/devhub/jordanstapinski/DEVHUB-185/learn)

This PR fixes some bugs with state on the learn page. Namely, the following points were addressed:
- Clears all state when the top `learn` link is clicked on the nav bar. (DEVHUB-185)
- Adds a query param to the URL based on which tab is clicked (articles, podcasts, etc.)
- Fixes a bug where the `select` component would not reset state while controlled by a parent if cleared
- Fixes a bug where pagination would change which filter settings were applied
- Adds an "enum" (actually an object since we are not on TS) for the different tabs on the learn page
- Fixes a bug where the URL would flicker with query params when certain filters were applied

Acceptance Criteria:
- Filters, Text filter, pagination, tabs should all still work as expected on the Learn page
- Clicking the `Learn` link on the top nav should reset state at any time
- Pagination should properly work with the same filters applied
- Clicking on an article from the learn page and then going back in the browser should keep filter settings

### DIFF
--- a/cypress/integration/learn.js
+++ b/cypress/integration/learn.js
@@ -31,7 +31,7 @@ describe('Learn Page', () => {
             cy.get('[data-test="card"]')
                 .first()
                 .should('contain', 'Working with MongoDB Transactions')
-                .click();
+                .click({ force: true });
         });
         cy.url().should('include', FIRST_ARTICLE_IN_ORDERING);
         // By targeting the hero banner we can be sure the navigation is done
--- a/cypress/integration/learn.js
+++ b/cypress/integration/learn.js
@@ -80,7 +80,7 @@ describe('Learn Page', () => {
         // Using trigger seems to work, but this should be looked into a bit more
         cy.contains('Atlas (26)').trigger('click');
         // The url should contain the filter value as a param
-        cy.url().should('include', '?products=Atlas');
+        cy.url().should('include', 'products=Atlas');
         // Check content
         cy.checkFirstCardInCardList('Coronavirus Map');
     });
--- a/gatsby-browser.js
+++ b/gatsby-browser.js
@@ -1,5 +1,10 @@
-exports.shouldUpdateScroll = ({ prevRouterProps, routerProps }) => {
-    const prevLocation = prevRouterProps && prevRouterProps.location.pathname;
-    const newLocation = routerProps && routerProps.location.pathname;
+import { removeTrailingSlash } from './src/utils/remove-trailing-slash';
+
+export const shouldUpdateScroll = ({ prevRouterProps, routerProps }) => {
+    const prevLocation =
+        prevRouterProps &&
+        removeTrailingSlash(prevRouterProps.location.pathname);
+    const newLocation =
+        routerProps && removeTrailingSlash(routerProps.location.pathname);
     return prevLocation !== newLocation;
 };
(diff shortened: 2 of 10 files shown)


gatsby-browser.js
```diff
@@ -1,5 +1,10 @@
-exports.shouldUpdateScroll = ({ prevRouterProps, routerProps }) => {
-    const prevLocation = prevRouterProps && prevRouterProps.location.pathname;
-    const newLocation = routerProps && routerProps.location.pathname;
+import { removeTrailingSlash } from './src/utils/remove-trailing-slash';
+
+export const shouldUpdateScroll = ({ prevRouterProps, routerProps }) => {
```
jestapinski, line 3: I noticed sometimes the page would jump to the top when a filter was applied. That was because sometimes there is a trailing slash initially which then gets removed when the filters are applied. This check now accounts for any potential trailing slashes

src/utils/query-string.js
```diff
@@ -1,7 +1,10 @@
 export function buildQueryString(params) {
-    const entries = Object.keys(params).map(
-        key => encodeURIComponent(key) + '=' + encodeURIComponent(params[key])
-    );
+    const entries = Object.keys(params)
+        .sort()
```
jestapinski, line 3: Sorting here prevents the url params from flashing by adding some more consistency

src/pages/learn.js
```diff
@@ -265,27 +267,47 @@ export default ({
...
+            } else {
+                delete filterValue['content'];
+            }
+            setFilterValue({ ...filterValue });
+        },
+        [filterValue]
+    );
+
+    const activeContentTab = useMemo(
+        () => filterValue['content'] || LearnPageTabs.all,
```
jestapinski, line 283: This now will take a query param into account on a render

cypress/integration/learn.js
```diff
@@ -31,7 +31,7 @@ describe('Learn Page', () => {
             cy.get('[data-test="card"]')
                 .first()
                 .should('contain', 'Working with MongoDB Transactions')
-                .click();
+                .click({ force: true });
```
jestapinski, line 34: There were some Cypress issues where elements could become detached from the DOM based on a race condition. Here this card should continue to exist so it is safe to force. The re-render seems to be just hydrating in fonts from what I could dig up.

tahiya-mongo:
Looking good to me. 

Questions:
- Do we want to remove the text and select filter queries from the url when we navigate to video and podcast tabs?
- This may be a product requirement, but would it make sense to reset the filter states on page refresh as well clicking the `Learn` button on the top nav bar?

jestapinski:
Thanks for taking a look!

For removing when going to other tabs, we are going to leave for now. The thinking is twofold:

1. If someone navigates to videos and back to articles, they want these filters to still apply
2. We want to add filters to everything at some point so having this logic in makes sense instead of trying to remove it for certain pieces of content only to re-add it later.

For refresh, we want the filters to still be applied. This is also because now these are query params so the link becomes more shareable. For now we just want the clearing action on clicking `Learn` up top.

Great points! Does this make sense?

tahiya-mongo (approved):
Makes sense. Thanks for the clarification!

jestapinski:
Thanks for taking a look Tahiya!

PR #559, opened 2020-11-30
Fix SSR Rehydration for Learn Page Tabs

jestapinski:
[Staging Link (learn page with Articles tab selected in query param)](https://docs-mongodbcom-staging.corp.mongodb.com/master/devhub/jordanstapinski/rehydrate/learn/?content=Articles)
[Prod Link (learn page with Articles tab selected in query param)](https://developer.mongodb.com/learn/?content=Articles)

There is an SSR rehydration issue with the tabs on the learn page where they do not re-render to update styles after being applied at build time. What happens is someone may link right to a specific tab (`?content=Articles` for example) but the `All` tab still looks active.

~The suggested pattern here is to wrap in a `ClientOnly` component, which will only render this on the client so there are no lingering styles applied from build-time. Any other ideas for SSR rehydration would be awesome but this seems the most straight-forward to me.~

To verify, check out the links, verify via query params the Articles tab should be selected. This should be broken in prod and fixed in the given staging link.


jestapinski:
Found out a better solution for this issue which is pretty small so no review really needed.

The component in question had functions which were not being re-evaluated.

PR #567, opened 2020-12-15
DEVHUB-322: Let captions float with images in figures

jestapinski:
[JIRA](https://jira.mongodb.org/browse/DEVHUB-322)

[Broken Article](https://docs-mongodbcom-staging.corp.mongodb.com/master/devhub/jordanstapinski/staging/article/schema-design-anti-pattern-bloated-documents)
[Fixed Article](https://docs-mongodbcom-staging.corp.mongodb.com/master/devhub/jordanstapinski/DEVHUB-322/article/schema-design-anti-pattern-bloated-documents)

This PR fixes a bug that has been around but not uncovered until now. The idea is that with a `figure` that has some `align` option and a caption, the caption does not float with the image. The solution is to have the entire block float together! I also took this chance to clean up some CSS in the `figure`

Check out the article in the staging link and the broken article link. In the latter you will see for some GIFs the caption looks out of place. This is fixed here in the staging link!

A/C:
- Captions float with images
- Figure is unaffected in any other way
- Image is unaffected in any other way

### DIFF
--- a/src/components/Figure.js
+++ b/src/components/Figure.js
@@ -2,20 +2,31 @@ import React from 'react';
 import styled from '@emotion/styled';
 import CaptionLegend from './CaptionLegend';
 import Image from './Image';
+import { getImageAlignmentStyle } from '../utils/get-image-alignment-style';
 import { getNestedValue } from '../utils/get-nested-value';
 import { size } from './dev-hub/theme';
 
 const Figure = styled('figure')`
     margin: 0 0 ${size.articleContent};
+    width: ${({ figWidth }) => figWidth};
+    ${({ alignStyle }) => alignStyle};
 `;
-export default ({ nodeData, ...rest }) => (
-    <Figure
-        className="figure"
-        style={{
-            width: getNestedValue(['options', 'figwidth'], nodeData) || 'auto',
-        }}
-    >
-        <Image nodeData={nodeData} captioned={!!nodeData.children.length} />
-        <CaptionLegend {...rest} nodeData={nodeData} />
-    </Figure>
-);
+
+const removeAlignOption = nodeData => delete nodeData['options']['align'];
+
+export default ({ nodeData, ...rest }) => {
+    const customAlign = getNestedValue(['options', 'align'], nodeData);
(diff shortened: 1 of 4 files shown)


src/components/Figure.js
```diff
@@ -2,20 +2,31 @@ import React from 'react';
...
 import { size } from './dev-hub/theme';
 
 const Figure = styled('figure')`
     margin: 0 0 ${size.articleContent};
+    width: ${({ figWidth }) => figWidth};
+    ${({ alignStyle }) => alignStyle};
 `;
-export default ({ nodeData, ...rest }) => (
-    <Figure
-        className="figure"
```
jestapinski, line 13: I don't think we are using this `className` but could be wrong

src/components/Figure.js
```diff
@@ -2,20 +2,31 @@ import React from 'react';
...
-    </Figure>
-);
+
+const removeAlignOption = nodeData => delete nodeData['options']['align'];
+
+export default ({ nodeData, ...rest }) => {
+    const customAlign = getNestedValue(['options', 'align'], nodeData);
+    if (customAlign) {
+        // Apply the alignment to the figure, no need to also apply to image
+        removeAlignOption(nodeData);
```
jestapinski, line 21: We don't need to pass this to image as well now. If we do it is fine but would rather just deal with alignment here.

src/components/Figure.js
sophstad: Yup, it's used [in docs-tools](https://github.com/mongodb/docs-tools/search?l=CSS&q=figure) but y'all don't depend on that!

src/components/Figure.js
jestapinski: Actually we still do, going to undo this change.

sophstad (approved):
Nice find and fix, will probably steal this for Snooty! Thanks Jordan!

jestapinski:
Thanks for the quick look! Found a bug where `center` would not work appropriately since that does need to apply to the image so just cleaned that up by removing the logic to remove align from nodeData. Updated staging links as well

PR #570, opened 2020-12-16
DEVHUB-291: Hover Card

jestapinski:
[JIRA](https://jira.mongodb.org/browse/DEVHUB-291)
[Figma](https://www.figma.com/proto/QuUf86imwSoPNYoxZX3az2/Academia---Student-Spotlights?node-id=508%3A379&scaling=min-zoom)
[Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/devhub/jordanstapinski/DEVHUB-291/) (check out the events image on the home page, this is now a Hover Card. This home page change is not part of the PR)

This PR adds a new unused component to the library: `HoverCard`. This component presents an image and on focus or hover, displays its children above an opaque backdrop.

PR #571, opened 2020-12-17
DEVHUB-292: Reusable Grid

jestapinski:
[JIRA](https://jira.mongodb.org/browse/DEVHUB-292)
[Figma](https://www.figma.com/proto/QuUf86imwSoPNYoxZX3az2/Academia---Student-Spotlights?node-id=508%3A1&scaling=min-zoom)
[Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/devhub/jordanstapinski/DEVHUB-292/) (check out the grid under the events section on the home page)

This PR adds a re-usable `Grid` component (mainly for the hover cards in #570 but can be used for other components). The goal is for the `Grid` to take a repeatable structure based on the number of columns and the layout defined by span of rows/cols per item that should be in the grid.

To help try and clarify how the grid is constructed and accessed, I abstracted the Grid position calculation to a class `GridLayout` which when given the rows/cols span blueprint can get the `position` of an element `i` and return in it terms of row and col span. This has been captured in a series of jest tests.

Furthermore, this Grid is also added into the storybook.

Screen Shot 2020-12-17 at 12 27 30 PM


jestapinski:
Will get another pass in later UAT

PR #629, opened 2021-02-01
DEVHUB-423: Make Reference internal links actually internal

jestapinski:
[JIRA](https://jira.mongodb.org/browse/DEVHUB-423)
[Staging](https://docs-mongodbcom-staging.corp.mongodb.com/master/devhub/jordanstapinski/DEVHUB-423/)

This PR makes an update to the `reference` directive for links in articles. If an author want to link to something within DevHub, then let's turn this into an internal link for a few reasons:
- Performance boost
- Better SEO since we then will tack on a trailing slash to avoid redirects (#628 )

To check, articles with links should continue to work as expected, except staging links will now redirect within staging


jestapinski:
I also added our aliases into the jest module mapping

PR #702, opened 2021-03-05
DEVHUB-514: Pull in MVP search

jestapinski:
[Staging](https://docs-mongodbcom-staging.corp.mongodb.com/CI/devhub/jordanstapinski/DEVHUB-514/)

By no means is this ready for production. This PR adds the base code for the Search UI from the Docs Platform. We will continue to clean unused fields as development continues, but this gives us the UI to work with. We also hooked it to the existing text filter backend at this time.

I had to upgrade the `icon` component when also adding the `IconButton`. I also explicitly added the `TextInput` dep from LeafyGreen. There are some other leafygreen deps we don't explicitly have in `package.json` at this time but we will add them in if we for-sure need them.


jestapinski:
Merging for now, since this is not going into staging and the e2e failure is due to overlapping on mobile. This will be addressed when we do mobile.

PR #716, opened 2021-03-11
DEVHUB-503: Move all instances of developer.mongodb.com to one variable in repo

Dmytro-Shustra-EPAM:
This PR adds constant for developer.mongodb.com and replaces places where this constant needed.


jestapinski:
can we also update in `gatsby-config.js` and `gatsby-config-prod.js`?

Dmytro-Shustra-EPAM:
> can we also update in `gatsby-config.js` and `gatsby-config-prod.js`?

Config is node.js part and for constans export should use module.exports. I have tried 'required' and 'import' in that files but the result had an error.

jestapinski (approved):
Looks good to me, I went ahead and updated the snapshots on my machine as my results were consistent with CI. I am on node 12.18.4, npm 6.14.6 and jest 26.6.3. Not sure if there is something else different about the environment but I would not worry for now.

Give it a try and let me know if there are still issues, this code though looks good to me so thanks!

PR #734, opened 2021-03-22
DEVHUB-549: Search on devhub with / shortcut

Dmytro-Shustra-EPAM:
This PR adds the opportunity to open a search bar on devhub with / (slash).

### DIFF
--- a/src/components/dev-hub/searchbar/Searchbar.js
+++ b/src/components/dev-hub/searchbar/Searchbar.js
@@ -7,6 +7,7 @@ import React, {
 } from 'react';
 import { css } from '@emotion/core';
 import styled from '@emotion/styled';
+import { useEventListener } from '@leafygreen-ui/hooks';
 import CondensedSearchbar from './CondensedSearchbar';
 import ExpandedSearchbar, {
     GoButton,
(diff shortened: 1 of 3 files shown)


src/components/dev-hub/searchbar/Searchbar.js
```diff
@@ -135,6 +137,15 @@ const Searchbar = ({ isExpanded, setIsExpanded }) => {
         }
         setIsFocused(true);
     }, [isFocused]);
+
+    // Focus when slash pressed
+    const onSlashKeyUp = useCallback(({ key, keyCode }) => {
+        (key === '/' || keyCode === SLASH_KEY) && onExpand();
+    }, [onExpand]);
```
jestapinski, line 144: This is great when the searchbar is condensed by default. However on screen widths above 1200px the searchbar is open by default and we would also like for it to focus there as well. Can we also include focusing that element here for the full size?

Dmytro-Shustra-EPAM: Thanks for comment. I have fixed it and also have deleted prop 'value' for ExpandedSearchbar because component doesn't expect it. Have some suggestion to check SearchTextInput. SearchTextInput has 'isSearching' prop for styled component (mobile version) but for current variant this prop always has false. Searchbar doesn't use isFocused for mobile ExpandedSearchbar so like result 'isSearching' always is false for SearchTextInput. We can ignore  if it's unnecessary or create a task for the future.

jestapinski: Great, thanks for finding. I have created a low priority ticket for us to tackle in the future for it. We pulled this component from another team so it is likely we didn't remove everything we didn't need to clean up, thanks!

jestapinski (approved):
Awesome, thanks Dmytro!

PR #735, opened 2021-03-22
DEVHUB-560: Implement page number into page title for numbered article pages

Dmytro-Shustra-EPAM:
This PR adds a new page title for numbered article pages also refreshes some tests.

### DIFF
--- a/src/templates/tag.js
+++ b/src/templates/tag.js
@@ -133,7 +133,7 @@ const Tag = props => {
 
             <ArticleContent>
                 {isAuthor && <H3>Articles by {name}</H3>}
-                <CardList articles={[...articles, ...articles, ...articles]} />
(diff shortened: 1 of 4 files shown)


src/templates/tag.js
```diff
@@ -3,6 +3,7 @@ import { Helmet } from 'react-helmet';
 import styled from '@emotion/styled';
 import { css } from '@emotion/core';
 import PropTypes from 'prop-types';
+import queryString from 'query-string';
```
jestapinski, line 6: Could we swap the `query-string` library use with the `parseQueryString` function in `src/utils/query-string.js` for both?

jestapinski (approved):
Awesome, thanks Dmytro! Feel free to merge when ready

PR #736, opened 2021-03-22
DEVHUB-561: Implement a H1 to all paginated article pages which contains the page number

Dmytro-Shustra-EPAM:
This PR changes H2 to H1 tags without style changes.

### DIFF
--- a/src/pages/learn.js
+++ b/src/pages/learn.js
@@ -3,7 +3,7 @@ import styled from '@emotion/styled';
 import { withPrefix } from 'gatsby';
 import { Helmet } from 'react-helmet';
 import Layout from '../components/dev-hub/layout';
-import { H1 } from '../components/dev-hub/text';
+import { H2 } from '../components/dev-hub/text';
 import MediaBlock from '../components/dev-hub/media-block';
 import Card from '../components/dev-hub/card';
 import CardList from '../components/dev-hub/card-list';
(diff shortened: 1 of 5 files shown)


src/pages/learn.js
```diff
@@ -70,6 +71,14 @@ const HeaderContent = styled('div')`
     max-width: ${FEATURED_ARTICLE_MAX_WIDTH};
     margin-left: auto;
     margin-right: auto;
+    ${H1} {
```
jestapinski, line 74: We can take a different approach that won't require us to re-do the H2 styling here.

Let's keep importing H2 from the text module, and we can use this line in the component to render it with an `h1` tag as such:

```
const H2asH1 = H2.withComponent('h1');
```

then when we render, we use this new tag so we can do:

```
<H2asH1>Make better, faster applications</H2asH1>
```

and similar for below, does this make sense?

src/templates/tag.js
```diff
@@ -112,7 +122,7 @@ const Tag = props => {
                             <AuthorByline>
                                 <SyledAuthorImage image={author_image} />
                                 <AuthorName>
-                                    <H2>{name}</H2>
+                                    <H1>{name}</H1>
```
jestapinski, line 125: We also want to add the pagination into here as well like we are doing for the `title` in the other PR. Can we add that in here please?

src/pages/learn.js
Dmytro-Shustra-EPAM: Yeah, looks better and clear, thanks for a tip.

src/templates/tag.js
```diff
@@ -117,7 +118,7 @@ const Tag = props => {
                             <AuthorByline>
                                 <SyledAuthorImage image={author_image} />
                                 <AuthorName>
-                                    <H2>{name}</H2>
+                                    <Title>{name}</Title>
```
jestapinski, line 121: Code functionally looks great, we just want to add one more piece here similar to #735 with the page number. This time we also want the page number in the H1, did we want to add it in here as well?

We would like something like this screenshot here, let me know if this does not make sense.
Screen Shot 2021-03-24 at 4 05 13 PM

Dmytro-Shustra-EPAM: Thanks for help. I have added pagination logic for h1 title, please check new changes.

jestapinski (approved):
Great, thanks!

PR #742, opened 2021-03-23
DEVHUB-549: Ensure slash key works on Firefox

jestapinski:
[Try it out here](https://docs-mongodborg-staging.corp.mongodb.com/master/devhub/jordanstapinski/DEVHUB-549-ff/)

I noticed the Slash did not work on Firefox because Firefox opens a quick find window with that key. We need to use `preventDefault` in order to block that. I also swapped the event from keyUp to keyDown so it would work as soon as the key has been pressed. I also shortened the UI test and was able to omit the `wait`

Give it a try with the link above!

### DIFF
--- a/cypress/integration/search.js
+++ b/cypress/integration/search.js
@@ -55,6 +55,15 @@ describe('search', () => {
         cy.visitWithoutFetch('/');
         checkCondensedSearchbar();
     });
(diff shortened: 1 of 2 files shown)


Sheeri:
Confirmed! LGTM :D

jestapinski:
Thanks for taking a look @Sheeri !

PR #749, opened 2021-03-25
DEVHUB-566: Fix Pop-up on Search Results - MongoDB

Dmytro-Shustra-EPAM:
This PR fixes pop-up on Search Results. The problem is in 'articles' and 'ActiveCardList' component. When we have some result from search bar - 'articles' will be empty, that why the scrollbar is blinking.


jestapinski (approved):
Awesome, thanks!

PR #757, opened 2021-04-01
DEVHUB-556: Create Podcast Pages

Dmytro-Shustra-EPAM:


src/utils/setup/create-podcast-pages.js
```diff
@@ -0,0 +1,21 @@
...
+export const createPodcastPages = async (createPage, metadataDocument) => {
+    const { allPodcasts } = await fetchBuildTimeMedia();
+
+    allPodcasts.forEach(podcast => {
+        const urlSuffix = getTagPageUriComponent(podcast.title);
+        createPage({
+            path: `/podcasts/${urlSuffix}`,
+            component: path.resolve(`./src/templates/podcast.js`),
+            context: {
+                snootyStitchId: SNOOTY_STITCH_ID,
```
jestapinski, line 15: We should be able to omit this field

src/utils/setup/fetch-build-time-media.js
```diff
@@ -3,10 +3,11 @@ import { STITCH_AUTH_APP_ID } from '../../constants';
 import { parsePodcasts } from '../parse-podcasts';
 import { simplifyTwitchResponse } from '../simplify-twitch-response';
 import { simplifyYoutubeResponse } from '../simplify-youtube-response';
+import memoizerific from 'memoizerific';
 
 const MAX_RESULTS = 5;
 
-export const fetchBuildTimeMedia = async () => {
+const fetch = async () => {
```
jestapinski, line 10:
```suggestion
const fetchContent = async () => {
```

Just want to be sure we don't overwrite the built-in `fetch` lib.

jestapinski (approved):
Great work, had two small improvements but this is good to merge when ready. Thanks Dymtro!

PR #760, opened 2021-04-05
DEVHUB-592: Create a Generalized Form

jestapinski:
This PR begins to add logic to support "Generalized Forms", which are forms that can be re-used with incoming data about the layout of a form.

There is still more work to be done, but this begins to allow us to render forms more easily.

We should be sure of the following:
- Form state works appropriately
- Form labels/placeholders work appropriately

We will add some tests for this when the full project is coming together, but for now we can continue to clean this up so we can keep moving on the project.

### DIFF
--- a/src/components/dev-hub/cms-form.js
+++ b/src/components/dev-hub/cms-form.js
@@ -1,0 +1,50 @@
+import React from 'react';
+import Checkbox from '@leafygreen-ui/checkbox';
+import Input from './input';
+import TextArea from './text-area';
+
+const FORM_ELEMENT_TYPES = {
+    CHECKBOXES: 'Checkboxes',
+    EMAILINPUT: 'EmailInput',
+    TEXTAREA: 'Textarea',
+};
+
+const mapTypeToFormElement = (type, labels) => {
+    switch (type) {
+        case FORM_ELEMENT_TYPES.CHECKBOXES:
+            return labels.map(({ label }) => (
+                <Checkbox key={label} bold darkMode={true} label={label} />
+            ));
+        case FORM_ELEMENT_TYPES.EMAILINPUT:
+            return (
+                <Input
+                    key={labels[0].label}
+                    bold
+                    darkMode={true}
+                    label={labels[0].label}
+                    type="email"
+                />
+            );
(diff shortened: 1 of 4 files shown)


src/components/dev-hub/cms-form.js
```diff
@@ -0,0 +1,90 @@
...
+                    label={label}
+                />
+            ));
+        case FORM_ELEMENT_TYPES.EMAILINPUT:
+            return (
+                <Input
+                    key={labels[0].label}
+                    bold
+                    darkMode={true}
+                    label={labels[0].label}
```
Dmytro-Shustra-EPAM, line 42: Nice work! Just have a suggestion, if it is possible could we change our data on server and create object 
{ label: 'Some Labe', placeholder: 'SomePlaceholder', ...another_props } instead of {labels: [{label}, {placeholder}]}?

src/components/dev-hub/cms-form.js
```diff
@@ -0,0 +1,90 @@
...
+
+const mapTypeToFormElement = (type, labels, state, dispatch, name) => {
+    switch (type) {
+        case FORM_ELEMENT_TYPES.CHECKBOXES:
+            return labels.map(({ label }, index) => (
+                <Checkbox
+                    key={`${name}-${index}`}
+                    bold
+                    onChange={e =>
+                        dispatch({
```
Dmytro-Shustra-EPAM, line 27: Looks like dispatch for checkbox always return 'on' even if checkbox unchecked. Try to check and after uncheck checkbox - e.target.value will be 'on' for two cases.

jestapinski: Good point, I fixed this to use the `checked` field instead of `value` so this should work as expected.

src/components/dev-hub/cms-form.js
jestapinski: Good question! I was able to add a new field for `placeholder` but cannot remove the object within the array for now. The Strapi CMS out of the box limits what types we can use in an array to be an object. This can be changed but for now I would like to leave as object since we can change in the future.

Dmytro-Shustra-EPAM (approved):
Nice work, thanks!

PR #761, opened 2021-04-06
DEBHUB-602: Star Rating

Dmytro-Shustra-EPAM:
This PR adds star rating component for article page. I would like to add this component to the article in another PR. Also, any tips for SVG and styles will helpful. 

Screenshot 2021-04-06 at 15 54 12
Screenshot 2021-04-06 at 15 53 07


src/components/dev-hub/star-rating.js
```diff
@@ -0,0 +1,180 @@
...
+            nextState = { ...nextState, third: true };
+        case 'second':
+            nextState = { ...nextState, second: true };
+        case 'first':
+            nextState = { ...nextState, first: true };
+            break;
+        case 'clicked':
+            nextState = { ...state, isClicked: true };
+            break;
+        case 'clear':
```
jestapinski, line 75: Could we use something similar to an `enum` for these so we can refer to them more easily? One pattern that may be helpful is using an object for keys/values and referring to keys:

```
const STAR_ACTIONS = {
  FIRST: 'first',
  SECOND: 'second',
  ...
  CLEAR: 'clear'
}
```

src/components/dev-hub/star-rating.js
```diff
@@ -0,0 +1,180 @@
...
+        case 'fifth':
+            nextState = { ...nextState, fifth: true };
+        case 'fourth':
+            nextState = { ...nextState, fourth: true };
+        case 'third':
+            nextState = { ...nextState, third: true };
+        case 'second':
+            nextState = { ...nextState, second: true };
+        case 'first':
+            nextState = { ...nextState, first: true };
```
jestapinski, line 70: I like this approach to adding in first, second, third, etc.! Nice work

A small performance optimization might be to only spread `nextState` at the end instead of (potentially) 5 times (one for each switch case). We could start a new object which we add to for making `fifth` `true`, and `fourth` `true`, etc.

Then at the end we would return a new object which spreads `nextState` and this object we incrementally build. Thoughts?

src/components/dev-hub/icons/star-icon.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { useTheme } from 'emotion-theming';
```
jestapinski, line 3: The package name for this has recently changed! I would recommend pulling in the latest from the branch and running `npm ci` to see it on your end.

```suggestion
import { useTheme } from '@emotion/react';
```

src/components/dev-hub/star-rating.js
```diff
@@ -0,0 +1,180 @@
...
+    );
+    const onLeaveHandler = useCallback(() => {
+        !state.isClicked && dispatch('clear');
+    }, [state.isClicked]);
+
+    return (
+        <StyledContainer>
+            <StyledTitle>Rate this article</StyledTitle>
+            <StyledStarsContainer onMouseLeave={onLeaveHandler}>
+                <StyledStar
```
jestapinski, line 112: Would is be possible to use a `button` tag as this is semantically a button? We have the `Button` component which should be easy to style without making it `primary` or `secondary`

That also allows us to get CSS like the `cursor` built-in already.

src/components/dev-hub/star-rating.js
jestapinski: We could also use an array as well to reduce the number of keys in our object and be able to scale to different numbers of stars.

src/components/dev-hub/star-rating.js
```diff
@@ -0,0 +1,180 @@
...
+                    isActive={state.first}
+                    isClicked={state.isClicked}
+                    name="first-star"
+                    onClick={() => onClickHandler(onFirstsClick)}
+                    onMouseEnter={() => onHoverHandler('first')}
+                    width={state.first && size.mediumLarge}
+                >
+                    <stop offset="0%" stopColor="#D34F94" />
+                    <stop offset="100%" stopColor="#D95B8F" />
```
jestapinski, line 121: Perhaps we could re-use these in a new component above and pass the start and end colors as props for the gradient? (Or even better, we could turn one full star into a reusable component)

src/components/dev-hub/star-rating.js
```diff
@@ -0,0 +1,180 @@
...
+        <StyledContainer>
+            <StyledTitle>Rate this article</StyledTitle>
+            <StyledStarsContainer onMouseLeave={onLeaveHandler}>
+                <StyledStar
+                    isActive={state.first}
+                    isClicked={state.isClicked}
+                    name="first-star"
+                    onClick={() => onClickHandler(onFirstsClick)}
+                    onMouseEnter={() => onHoverHandler('first')}
+                    width={state.first && size.mediumLarge}
```
jestapinski, line 118: I noticed when we increase the size it seems to push to the right. We would like the star to stay centered in its bounding box but expand from the center. If we hover over it we can see this happening:

Screen Shot 2021-04-06 at 10 37 13 AM
Screen Shot 2021-04-06 at 10 37 20 AM

src/components/dev-hub/star-rating.js
```diff
@@ -0,0 +1,180 @@
+import React, { useCallback, useReducer } from 'react';
+import PropTypes from 'prop-types';
+import StarIcon from '~components/dev-hub/icons/star-icon';
+import styled from '@emotion/styled';
+import { size, fontSize, screenSize } from './theme';
+
+const StyledContainer = styled.div`
```
jestapinski, line 7:
```suggestion
const StyledContainer = styled('div')`
```

Let's keep consistent by using `()` notation for these

src/components/dev-hub/star-rating.js
```diff
@@ -0,0 +1,180 @@
...
+    display: flex;
+    flex-direction: column;
+
+    @media ${screenSize.mediumAndUp} {
+        flex-direction: row;
+    }
+`;
+
+const StyledStar = styled(StarIcon)`
+    width: ${size.mediumLarge};
```
jestapinski, line 17: Let's also alphabetize our CSS please

jestapinski (commented):
Awesome start Dmytro! I had some comments on the functionality and code style

Dmytro-Shustra-EPAM:
Thanks for comments! I will try to improve PR later.

src/components/dev-hub/star-rating.js
Dmytro-Shustra-EPAM: I would like to keep this variant. If we need some manipulation with colors this variant allow change   amount of colors and set offset for them.

src/components/dev-hub/star-rating.js
```diff
@@ -0,0 +1,213 @@
...
+
+    @media ${screenSize.mediumAndUp} {
+        margin: ${({ isActive }) => (isActive ? '0' : '0 4px')};
+        padding: 4px;
+        svg {
+            height: auto;
+            width: auto;
+        }
+    }
+`;
```
jestapinski, line 30: One nitpick I have is to keep the CSS alphabetized, could we please reorder some of these properties? (Leaving media queries at the end is good)? Thanks!

jestapinski (approved):
Great work, tried it out myself and looks awesome. Great job!

src/components/dev-hub/star-rating.js
Dmytro-Shustra-EPAM: Thanks for help! I will fix it and merge PR after checking.

PR #763, opened 2021-04-09
DEBHUB-610: Implement Rating to Article page

Dmytro-Shustra-EPAM:
https://jira.mongodb.org/browse/DEVHUB-610

This PR implements component to article page. CSS was changed and now we have 2 variants - bottom rating (< 767px and > 767px) and top rating (< 767px and < 1023px and > 1023px). Check screenshots belowe.

Screenshot 2021-04-09 at 17 41 06
Screenshot 2021-04-09 at 17 41 27
Screenshot 2021-04-09 at 17 41 44
Screenshot 2021-04-09 at 17 41 50
Screenshot 2021-04-09 at 17 42 08

### DIFF
--- a/src/components/dev-hub/star-rating.js
+++ b/src/components/dev-hub/star-rating.js
@@ -29,7 +29,7 @@ const StyledButton = styled(Button)`
     }
 `;
 
-const StyledContainer = styled('div')`
+export const StyledContainer = styled('div')`
     display: flex;
     flex-direction: column;
 
@@ -53,7 +53,7 @@ const StyledTitle = styled('span')`
     }
 `;
 
-const StyledStarsContainer = styled('div')`
+export const StyledStarsContainer = styled('div')`
     align-items: center;
     display: flex;
 
--- a/src/templates/article.js
+++ b/src/templates/article.js
@@ -23,7 +23,10 @@ import { getNestedValue } from '../utils/get-nested-value';
 import { findSectionHeadings } from '../utils/find-section-headings';
 import { getNestedText } from '../utils/get-nested-text';
 import ArticleSchema from '../components/dev-hub/article-schema';
-import ArticleRating from '~components/ArticleRating';
+import {
+    default as ArticleRating,
+    StyledContainerTop,
+} from '~components/ArticleRating';
 
 /**
  * Name map of directives we want to display in an article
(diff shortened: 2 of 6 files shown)


src/templates/article.js
```diff
@@ -200,20 +232,24 @@ const Article = props => {
...
+                            width={size.default}
+                        />
+                        <ShareMenu
+                            title={articleTitle}
+                            url={articleUrl}
+                            height={size.default}
+                            width={size.default}
+                        />
+                    </MenuContainer>
+                    <ArticleRating isTop />
```
jestapinski, line 249: Did we lose the highlight effect for the stars? They now blend into the background on a hover

Screen Shot 2021-04-14 at 2 03 54 PM

src/templates/article.js
```diff
@@ -78,29 +81,58 @@ const ArticleContent = styled('article')`
...
+    display: flex;
     margin: ${size.tiny} ${size.default};
     span {
         padding: 0 ${size.tiny};
     }
+
     @media ${screenSize.largeAndUp} {
-        display: flex;
         flex-direction: column;
+        margin-top: 80px;
```
jestapinski, line 108: This seems a bit hacky to me. Could we have the `ArticleRating` outside of `Icons` and then below have content as normal (aligned next to each other)? Or, we can move outside of `Container` as then the flex won't apply.

CSS Grid might be a good approach here because we can easily change the grid layout for mobile by using something like `grid-template-areas`. Check out this link https://developer.mozilla.org/en-US/docs/Web/CSS/grid-template-areas

src/templates/article.js
```diff
@@ -78,29 +81,58 @@ const ArticleContent = styled('article')`
     max-width: ${size.maxContentWidth};
     padding-left: ${size.small};
     padding-right: ${size.small};
+
+    ${StyledContainerTop} {
```
jestapinski, line 85: We can clean up the styling a bit by doing something like `const StyledArticleContent = styled(ArticleContent)` and then applying style there. We can do this by adding a new prop to `ArticleContent` called `className` and passing that to its rendered component.

[Feel free to see the documentation for this](https://emotion.sh/docs/styled#styling-any-component) and let me know if there are any problems!

src/components/dev-hub/star-rating.js
```diff
@@ -29,7 +29,7 @@ const StyledButton = styled(Button)`
     }
 `;
 
-const StyledContainer = styled('div')`
+export const StyledContainer = styled('div')`
```
jestapinski, line 32: Using the `className` feature in the other comment we likely won't need to export these anymore

jestapinski (commented):
Good start! Some comments mainly on the layout, let me know what you think and thank you!

PR #770, opened 2021-04-13
DEVHUB-594 (prep): Add Endpoints for Form Submission

jestapinski:
This PR adds three functions which can be used in the devhub feedback project for handling actions with the modal. They are as follows:

- `createFeedback` which is to be called when the modal is opened. This creates an object in our DB for handling the user's feedback and returns that id
- `updateFeedback` which is to be called as the user enters more data
- `submitFeedback` which is to be called on completion

These are hosted in our backend-as-a-service on MongoDB Realm. Let me know if it is unclear how we plan to use these, thanks!

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -30,7 +30,9 @@ export const ADMONITIONS = [
     'warning',
 ];
 
+// Realm App IDs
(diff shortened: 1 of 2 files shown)


Dmytro-Shustra-EPAM (approved):

jestapinski:
Thanks!

PR #801, opened 2021-05-07
DEVHUB-643: Add Social Share Tags to Landing Pages

Dmytro-Shustra-EPAM:
This PR adds additional meta information for SEO.    
https://jira.mongodb.org/browse/DEVHUB-643

### DIFF
--- a/src/components/dev-hub/page-helmet.js
+++ b/src/components/dev-hub/page-helmet.js
@@ -15,8 +15,6 @@ const PageHelmet = ({ canonicalUrl, metaDescription, metaTitle }) => {
     const title = metaTitle || standardTitle;
     const url = canonicalUrl || siteUrl;
 
-    console.log(standardTitle);
-
     return (
         <SEO
             articleTitle={title}
--- a/src/components/dev-hub/page-helmet.js
+++ b/src/components/dev-hub/page-helmet.js
@@ -1,0 +1,45 @@
+import React, { memo } from 'react';
+import PropTypes from 'prop-types';
+
+import SEO from '~components/dev-hub/SEO';
+import { useSiteMetadata } from '~hooks/use-site-metadata';
+
+const STANDARD_DESCRIPTION =
+    "Code, content, tutorials, programs and community to enable developers of all skill levels on the MongoDB Data Platform which includes Atlas, Realm, Compass, Data Lake and more. Whether you're coding in Java, JavaScript, C#, Python, Node, Go or looking for how this fits with IOT, AI, ML - join or follow us here.";
+
+const PageHelmet = ({ canonicalUrl, metaDescription, metaTitle }) => {
+    const { title: standardTitle, siteUrl } = useSiteMetadata();
+
+    const description = metaDescription || STANDARD_DESCRIPTION;
+    const image = `${siteUrl}/public/images/MongoDB_Leaf.svg`;
+    const title = metaTitle || standardTitle;
+    const url = canonicalUrl || siteUrl;
+
+    console.log(standardTitle);
+
+    return (
+        <SEO
+            articleTitle={title}
+            canonicalUrl={url}
+            image={image}
+            metaDescription={description}
+            ogDescription={description}
+            ogTitle={title}
+            ogUrl={url}
+            type="website"
+            twitter={{
+                description,
+                image,
+                title,
+            }}
+        />
+    );
+};
+
+PageHelmet.propTypes = {
+    canonicalUrl: PropTypes.string,
+    metaDescription: PropTypes.string,
+    metaTitle: PropTypes.string,
+};
+
+export default memo(PageHelmet);
--- a/src/pages/academia/educators.js
+++ b/src/pages/academia/educators.js
@@ -1,21 +1,26 @@
 import React from 'react';
 import Layout from '../../components/dev-hub/layout';
-import { Helmet } from 'react-helmet';
 import { useSiteMetadata } from '../../hooks/use-site-metadata';
 import {
     TopBanner,
     HowToJoin,
     ProgramBenefits,
 } from '../../components/pages/educators';
+import PageHelmet from '~components/dev-hub/page-helmet';
+
+const PAGE_DESCRIPTION =
+    'MongoDB for Academia is for educators who want to prepare students for careers that require in-demand database skills that power modern applications.';
 
 export default () => {
-    const metadata = useSiteMetadata();
+    const { title, siteUrl } = useSiteMetadata();
     return (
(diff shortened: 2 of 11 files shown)


jestapinski:
Let's look at the CI tests and then I will take a look at the code, seems like we may have an extra slash in places.

Dmytro-Shustra-EPAM:
> Let's look at the CI tests and then I will take a look at the code, seems like we may have an extra slash in places.

Yeah, have checked already. Just forgot about to add slash to some URLs.

src/components/dev-hub/page-helmet.js
```diff
@@ -0,0 +1,43 @@
...
+import { useSiteMetadata } from '~hooks/use-site-metadata';
+
+const STANDARD_DESCRIPTION =
+    "Code, content, tutorials, programs and community to enable developers of all skill levels on the MongoDB Data Platform which includes Atlas, Realm, Compass, Data Lake and more. Whether you're coding in Java, JavaScript, C#, Python, Node, Go or looking for how this fits with IOT, AI, ML - join or follow us here.";
+
+const PageHelmet = ({ canonicalUrl, metaDescription, metaTitle }) => {
+    const { title: standardTitle, siteUrl } = useSiteMetadata();
+
+    const description = metaDescription || STANDARD_DESCRIPTION;
+    const image = `${siteUrl}/public/images/MongoDB_Leaf.svg`;
```
jestapinski, line 14: Can we make this a prop with a default value? Different pages may wish to supply this, but I like the idea of a default.

src/pages/community/events.js
```diff
@@ -23,26 +23,27 @@ const breadcrumbs = [
     { label: 'Events', target: '/community/events' },
 ];
 
+const PAGE_DESCRIPTION =
+    'Join us in a city near you to connect with MongoDB users who are shaking up their industries, and be among the first to hear updates and exciting news.';
+
 export default () => {
     const [events, error, isLoading] = useEventData();
+    const { title, siteUtl } = useSiteMetadata();
```
jestapinski, line 31:
```suggestion
    const { title, siteUrl } = useSiteMetadata();
```

Looks like this is a typo

src/components/dev-hub/page-helmet.js
```diff
@@ -0,0 +1,43 @@
+import React, { memo } from 'react';
+import PropTypes from 'prop-types';
+
+import SEO from '~components/dev-hub/SEO';
+import { useSiteMetadata } from '~hooks/use-site-metadata';
+
+const STANDARD_DESCRIPTION =
+    "Code, content, tutorials, programs and community to enable developers of all skill levels on the MongoDB Data Platform which includes Atlas, Realm, Compass, Data Lake and more. Whether you're coding in Java, JavaScript, C#, Python, Node, Go or looking for how this fits with IOT, AI, ML - join or follow us here.";
+
+const PageHelmet = ({ canonicalUrl, metaDescription, metaTitle }) => {
```
jestapinski, line 10: I would also like for us to optionally accept other SEO props as props here. The reason is soon we will be exposing these pages for changes via the CMS and we would like for the team to be able to adjust these if need be.

src/pages/academia/educators.js
```diff
@@ -1,21 +1,26 @@
...
-    const metadata = useSiteMetadata();
+    const { title, siteUrl } = useSiteMetadata();
     return (
         <Layout>
-            <Helmet>
-                <title>Academia for Educators - {metadata.title}</title>
-            </Helmet>
-            <TopBanner />
+            <PageHelmet
+                canonicalUrl={`${siteUrl}/academia/educators/`}
```
jestapinski, line 19: I see most of these use `siteUrl` for the canonical url which is correct. Would a better set of props for `PageHelmet` include just using the page path and then adding the `siteUrl` within `PageHelmet` to avoid the need to use it in every component which has `PageHelmet` as a child?

jestapinski: A similar thought is that many `title`s use the site title as well. Should we also append this automatically in `PageHelmet`?

src/components/dev-hub/page-helmet.js
```diff
@@ -0,0 +1,43 @@
...
+
+    return (
+        <SEO
+            articleTitle={title}
+            canonicalUrl={url}
+            image={image}
+            metaDescription={description}
+            ogDescription={description}
+            ogTitle={title}
+            ogUrl={url}
```
jestapinski, line 26: We also want to make sure the canonical and og urls for DevHub have a trailing slash. There is a util function we have called `addTrailingSlashIfMissing` which should be applied here.

jestapinski (commented):
Good so far! Some questions on props for the PageHelmet

src/pages/academia/educators.js
jestapinski: Okay sounds good to me if we leave `title` more generic here and continue to use from the calling component, that works for me.

Similar thought with the `siteUrl`. I don't think we would need to replace this for any of the main pages.

src/pages/community/events.js
```diff
@@ -28,14 +28,14 @@ const PAGE_DESCRIPTION =
...
+    const { title } = useSiteMetadata();
 
     return (
         <Layout>
             <PageHelmet
-                metaTitle={`Events - ${title}`}
-                canonicalUrl={`${siteUtl}/community/events/`}
-                metaDescription={PAGE_DESCRIPTION}
+                title={`Events - ${title}`}
+                pagePath="/community/events"
```
jestapinski, line 37: For `pagePath` Gatsby gives this component `props`, one of which is `uri` and another is `path` which might be useful here so we don't have to hardcode. If the pagePath is the `uri` we should use that. What do you think?

src/templates/tag.js
```diff
@@ -83,8 +83,8 @@ const Tag = props => {
     return (
         <Layout>
             <PageHelmet
-                canonicalUrl={`${metadata.siteUrl}${slug}/`}
-                metaTitle={metaTitle}
+                pagePath={`${slug}`}
```
jestapinski, line 86: Would `pagePath={slug}` also work? (Shorter syntax)

jestapinski (commented):
A few cleanup comments, but looks pretty good overall

src/pages/community/events.js
Dmytro-Shustra-EPAM: Thanks! Nice suggestion I will add.

jestapinski (approved):
Awesome work, this is great! I am waiting for the product/design team to provide another image which we will use as the OG image and such. Once I get that I will update this PR and merge.

Thanks!

PR #811, opened 2021-05-12
DEVHUB-557: Swap Podcast functionality on Learn Page

Dmytro-Shustra-EPAM:
This PR adds new logic for podcasts on learn page. Now the click event opens a new page instead of player.


jestapinski (approved):
Awesome work! Looks great, and thanks for removing the old behavior.

PR #816, opened 2021-05-13
DEVHUB-661: Adds interfaces and classes for video content

Dmytro-Shustra-EPAM:
This PR adds typescript integration to create a `Video` interface and two classes `YoutubeVideo` and `TwitchVideo` which use the video interface to create a common representation of a video.


src/utils/simplify-twitch-response.ts
```diff
@@ -0,0 +1,5 @@
+import { TwitchResponse } from '../interfaces/responses/twitch-response';
+import { TwitchVideo } from '../classes/twitch-video';
+
+export const simplifyTwitchResponse = (video: TwitchResponse): TwitchVideo =>
```
jestapinski, line 4: This is a bit of a nitpick, but we can likely rename this function and the youtube one since they no longer really "simplify" anything but they map the response to a class instance. Do you think we should update the naming?

src/utils/simplify-youtube-response.ts
```diff
@@ -0,0 +1,13 @@
...
+import { YoutubeVideo } from '../classes/youtube-video';
+
+export const simplifyYoutubeResponse = (
+    responseData: YoutubeResponse
+): YoutubeVideo => {
+    const video = responseData.snippet;
+    // Video publish date is stored in "contentDetails", publish date in "snippet"
+    // is for when this video was added to the playlist
+
+    const publishedDate = responseData.contentDetails?.videoPublishedAt;
```
jestapinski, line 11: Let's move this logic into the `class` definition for YouTube. This is part of youtube-specific logic we can reuse every time we want to construct the class and so it should belong there.

src/interfaces/responses/twitch-response.ts
```diff
@@ -0,0 +1,19 @@
+export interface TwitchResponse {
+    id: string;
+    stream_id: string;
+    user_id: string;
```
jestapinski, line 4: Can we please alphabetize this list?

src/interfaces/responses/twitch-response.ts
```diff
@@ -0,0 +1,19 @@
...
+    created_at: string;
+    description: string;
```
jestapinski, line 6:
```suggestion
    description?: string;
```

Should we mark this as optional since some have no description? Or do they return empty string so it is okay if they are not optional?

src/interfaces/video.ts
```diff
@@ -0,0 +1,11 @@
+import { VideoType } from '../types/video-type';
+
+export interface Video {
+    description: string;
+    mediaType: VideoType;
+    publishDate: string;
+    thumbnailUrl: string;
+    title: string;
+    videoId: string;
+    playlistId?: string;
```
jestapinski, line 10: We may not even need this field, I checked the repo and did not see it used anywhere. Do you think we can remove this one?

jestapinski (commented):
Excellent use of typescript, this makes the videos much more organized! I had some small comments so curious what you think

src/utils/simplify-twitch-response.ts
Dmytro-Shustra-EPAM: Yeah, the naming can confuse, I will try to change the function name.

src/interfaces/video.ts
Dmytro-Shustra-EPAM: Nice catch, I will remove this property here and from classes but let keep for response interfaces.

jestapinski (approved):
Awesome work, thanks! Types/classes/interfaces look great

PR #817, opened 2021-05-13
Use Component Factory to compose styles

jestapinski:
[Staging](https://docs-mongodborg-staging.corp.mongodb.com/CI/devhub/jordanstapinski/use-factory-for-composition/)

This PR makes an improvement to how we render `emphasis` (italics), bold, and links. Previously, we assumed the initial child of this node was text. However, this may be a link, or some other node. As a result, we want to pass the node's children again through the component factory to compose all needed styles.

### DIFF
--- a/src/components/Emphasis.js
+++ b/src/components/Emphasis.js
@@ -1,8 +1,12 @@
 import React from 'react';
-import { getNestedValue } from '../utils/get-nested-value';
+import ComponentFactory from './ComponentFactory';
 
 const Emphasis = ({ nodeData }) => (
-    <em>{getNestedValue(['children', 0, 'value'], nodeData)}</em>
+    <em>
+        {nodeData.children.map((child, index) => (
+            <ComponentFactory nodeData={child} key={index} />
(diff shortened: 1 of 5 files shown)


src/components/Emphasis.js
```diff
@@ -1,8 +1,12 @@
 import React from 'react';
-import { getNestedValue } from '../utils/get-nested-value';
+import ComponentFactory from './ComponentFactory';
 
 const Emphasis = ({ nodeData }) => (
-    <em>{getNestedValue(['children', 0, 'value'], nodeData)}</em>
+    <em>
+        {nodeData.children.map((child, index) => (
+            <ComponentFactory nodeData={child} key={index} />
```
jestapinski, line 7: `ComponentFactory` gives the same behavior if the child is of type `text`. This continues to keep the existing functionality for Snooty and Strapi.

jestapinski:
I also ran `sitediff`, the only difference being the addition of composed styles in one Strapi article.

PR #820, opened 2021-05-13
DEVHUB-651: Add Table, TableRow, and TableCell support for Strapi

jestapinski:
[Staging (article with table)](https://docs-mongodbcom-staging.corp.mongodb.com/CI/devhub/jordanstapinski/DEVHUB-651/how-to/realm-test/#fourth-heading)
[JIRA](https://jira.mongodb.org/browse/DEVHUB-651)

This PR adds support for three new directives which are Strapi-exclusive to use tables. The Snooty `ListTable` concept really doesn't map well to the Markdown concept of a Table. Also, these tables support embedding images, inline code, etc!

This is not in production use yet, the above article is just one for testing Strapi rendering.

### DIFF
--- a/src/components/Table.js
+++ b/src/components/Table.js
@@ -1,0 +1,61 @@
+import React from 'react';
+import ComponentFactory from './ComponentFactory';
+import { getNestedValue } from '../utils/get-nested-value';
+import styled from '@emotion/styled';
+import { colorMap, size } from '../components/dev-hub/theme';
+
+const StyledTable = styled('table')`
+    border: 1px solid ${colorMap.greyLightTwo};
+    border-collapse: collapse;
+    margin: ${size.default} 0;
+    table-layout: fixed;
+    th,
+    td {
+        padding: ${size.small};
+        border: 1px solid ${colorMap.greyLightTwo};
+    }
+`;
+
+const TableContainer = styled('div')`
+    overflow-x: auto;
+`;
+
+const AlignedTH = styled('th')`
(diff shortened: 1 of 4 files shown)


src/components/Table.js
```diff
@@ -0,0 +1,61 @@
...
+                    />
+                </AlignedTH>
+            ))}
+        </tr>
+    </thead>
+);
+
+const Table = ({ nodeData: { align, children } }) => {
+    if (children && children.length) {
+        const headingRow = children[0] && children[0].children;
```
Dmytro-Shustra-EPAM, line 43: We can use the check operator for the short variant.
```suggestion
    if (children?.length) {
        const headingRow = children[0]?.children;
```

src/components/Table.js
```diff
@@ -0,0 +1,61 @@
+import React from 'react';
+import ComponentFactory from './ComponentFactory';
+import { getNestedValue } from '../utils/get-nested-value';
+import styled from '@emotion/styled';
+import { colorMap, size } from '../components/dev-hub/theme';
```
Dmytro-Shustra-EPAM, line 5: Can we use shortened import here? 
```suggestion
import { getNestedValue } from '~utils/get-nested-value';
import styled from '@emotion/styled';
import { colorMap, size } from '~components/dev-hub/theme';
```

Dmytro-Shustra-EPAM (approved):
Awesome! I just left some suggestions, but they can be omitted. Thanks!

jestapinski:
Thanks Dmytro!

PR #823, opened 2021-05-14
Podcasts: Add Exposed Links for Social Sharing

jestapinski:
[Staging Link](https://docs-mongodborg-staging.corp.mongodb.com/master/devhub/jordanstapinski/add-exposed-links/)

This PR mimics #804 to expose the links on the sides for the podcast pages. I also made sure to keep the tooltip text customizable. This way, if this project is ready for production first it can be deployed without the need for the devhub feedback project to be ready.

### DIFF
--- a/src/components/dev-hub/controlled-tooltip.js
+++ b/src/components/dev-hub/controlled-tooltip.js
@@ -4,7 +4,10 @@ import { css } from '@emotion/react';
 import Popover from 'react-tiny-popover';
 import { animationSpeed, layer, size } from './theme';
 
-const ControlledTooltipWrapper = styled('span')`
+const ControlledTooltipWrapper = styled('div')`
+    display: inline-block;
+    line-height: ${size.default};
+    height: ${size.default};
     /* Article sets these values */
     padding: 0 !important;
 `;
@@ -211,8 +214,11 @@ const Content = styled('div')`
         getArrowStyles(hasGradientBorder, position, theme)}
 `;
 
-const Trigger = styled('span')`
+const Trigger = styled('div')`
     cursor: pointer;
+    line-height: ${size.default};
+    display: inline-block;
+    height: ${size.default};
 `;
 
 // This logic was pulled from 'react-tiny-popover' (see = https://github.com/alexkatz/react-tiny-popover/blob/3928cfa67a57676f32d5f04b3e1decb8c31db544/src/Popover.tsx#L316-L352)
@@ -279,6 +285,7 @@ const ControlledTooltip = ({
     contentStyle,
     isOpen,
     setIsOpen,
+    ...props
 }) => {
     const ref = useRef(null);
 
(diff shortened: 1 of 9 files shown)


src/components/dev-hub/controlled-tooltip.js
```diff
@@ -4,7 +4,10 @@ import { css } from '@emotion/react';
 import Popover from 'react-tiny-popover';
 import { animationSpeed, layer, size } from './theme';
 
-const ControlledTooltipWrapper = styled('span')`
```
Dmytro-Shustra-EPAM, line 7: Screenshot 2021-05-17 at 23 43 28
Some components are using `span`  for CSS nesting and add margins let's move this logic here.

src/templates/podcast.js
```diff
@@ -83,6 +80,29 @@ const StyledPlayer = styled(AudioPlayer)`
...
+    align-items: center;
+    > :not(:first-child) {
+        margin-top: ${size.medium};
+    }
+    @media ${screenSize.upToLarge} {
+        display: inline-flex;
+        flex-direction: row;
+        > * {
+            margin-top: 0;
+            margin-left: ${size.mediumLarge};
```
Dmytro-Shustra-EPAM, line 94: Screenshot 2021-05-17 at 23 53 54
Looks like we don't need `margin-left` for the first element, let's omit first element.

src/templates/podcast.js
```diff
@@ -83,6 +80,29 @@ const StyledPlayer = styled(AudioPlayer)`
     }
 `;
 
+const StyledBlogShareLinks = styled(BlogShareLinks)`
+    flex-direction: column;
+    align-items: center;
+    > :not(:first-child) {
+        margin-top: ${size.medium};
```
Dmytro-Shustra-EPAM, line 87: Screenshot 2021-05-17 at 23 58 28
Also, here we need this margins only for screen 1024px and hire.

src/templates/podcast.js
```diff
@@ -83,6 +80,29 @@ const StyledPlayer = styled(AudioPlayer)`
...
+        > * {
+            margin-top: 0;
+            margin-left: ${size.mediumLarge};
+        }
+    }
+`;
+
+const StyledFooter = styled(ShareFooter)`
+    /* Target the copy link, although it is below some divs */
+    div > div > a {
```
Dmytro-Shustra-EPAM, line 101: Like nesting in CSS it's good, but we can import the styled components and use nesting with them, I prefer this way, but you can omit this suggestion.

Dmytro-Shustra-EPAM (commented):
Good so far! Some CSS is not exact for sharing links at the top (article, podcast). I have left comments, let's check this out icons position and responsive.

Dmytro-Shustra-EPAM (approved):
Awesome work! Thanks, @jestapinski!

jestapinski:
Thanks for the feedback @Dmytro-Shustra-EPAM !

PR #828, opened 2021-05-18
DEVHUB-663: Replace Video Functionality on the Learn Page

Dmytro-Shustra-EPAM:
[DEVHUB-663](https://jira.mongodb.org/browse/DEVHUB-663). This PR replaces Video Functionality on the Learn Page. Also is trying using TS for gatsby page generation and has test component that will be replaced  after [DEVHUB-622](https://jira.mongodb.org/browse/DEVHUB-662). For learn page was removed the hook which fetches new video data, so let’s be sure we will have all data during Gatsby generation.


src/templates/video.tsx
```diff
@@ -0,0 +1,15 @@
...
+type VideoProps = {
+    pageContext: {
+        data: IVideo;
+    };
+};
+
+const Video = ({ pageContext: { data: video } }: VideoProps) => {
+    console.log(video);
+
+    return <>666 TEST</>;
```
jestapinski, line 13: We should just return null or something else instead here for now.

src/utils/setup/fetch-build-time-media.js
```diff
@@ -3,10 +3,11 @@ import { STITCH_AUTH_APP_ID } from '../../constants';
 import { parsePodcasts } from '../parse-podcasts';
 import { transformTwitchResponse } from '../transform-twitch-response';
 import { transformYoutubeResponse } from '../transform-youtube-response';
+import memoizerific from 'memoizerific';
 
 const MAX_RESULTS = 5;
 
-export const fetchBuildTimeMedia = async () => {
+const fetchMedia = async () => {
```
jestapinski, line 10: Ah line 13 is where I believe we get only a few of the videos

src/utils/setup/fetch-build-time-media.js
```diff
@@ -23,3 +24,6 @@ export const fetchBuildTimeMedia = async () => {
         fallbackTwitchVideo: allTwitchVideos[0],
     };
 };
+export const fetchBuildTimeMedia = memoizerific(1)(
```
jestapinski, line 27: Great use of memoizerific here to reduce network calls

src/utils/get-tag-page-uri-component.js
```diff
@@ -5,7 +5,10 @@ const getTagPageUriComponent = tagPageUri => {
         case 'c#':
             return 'csharp';
         default:
-            return tagPageUri.toLowerCase().replace(/\W/g, '-');
+            return tagPageUri
+                .toLowerCase()
+                .replace(/\W/g, '-')
+                .replace(/--/g, '-');
```
jestapinski, line 11: Were we running into an issue with `--` before? Also not sure if this is working as expected since `/video/youtube/schema-design-anti-patterns--part-3/` still has a `--`.

src/utils/setup/create-video-pages.ts
```diff
@@ -0,0 +1,21 @@
+import path from 'path';
+import { fetchBuildTimeMedia } from './fetch-build-time-media';
+import { Video } from '../../interfaces/video';
+
+export const createVideoPages = async (
+    createPage: Function,
+    metadataDocument: object
```
jestapinski, line 7:
```suggestion
    metadata: object
```

If we rename here then on line 16 we can use shorthand to be `metadata` only.

jestapinski (commented):
Looks good to start! I had some comments

src/utils/get-tag-page-uri-component.js
Dmytro-Shustra-EPAM: Oh, for some titles with '-' (for example Anti-Patterns - Part 3) this regex is not enough, 
I will improve to better one `/[-_,]+/g`, for this regex it doesn’t matter how many '-' we will have.

src/templates/video.tsx
Dmytro-Shustra-EPAM: Yeah, but for final state of PR this line will disappear. I will replace this by component form another PR.

src/utils/setup/fetch-build-time-media.js
Dmytro-Shustra-EPAM: Nice catch! This is reason why we had not all pages!

src/utils/setup/create-video-pages.ts
```diff
@@ -0,0 +1,21 @@
...
+import { Video } from '../../interfaces/video';
+
+export const createVideoPages = async (
+    createPage: Function,
+    metadata: object
+) => {
+    const { allVideos } = await fetchBuildTimeMedia();
+
+    allVideos.forEach((video: Video) => {
+        createPage({
```
jestapinski, line 12: Oh nice! Easy enough to construct these pages. Would we like to do feedback on the pages themselves in this PR or should we come back to it after the PR?

src/utils/get-tag-page-uri-component.js
```diff
@@ -5,7 +5,10 @@ const getTagPageUriComponent = tagPageUri => {
         case 'c#':
             return 'csharp';
         default:
-            return tagPageUri.toLowerCase().replace(/\W/g, '-');
+            return tagPageUri
+                .toLowerCase()
+                .replace(/\W/g, '-')
+                .replace(/[-_,]+/g, '-')
```
jestapinski, line 11: 👍

jestapinski (approved):
Looks fantastic, and with data in the pages! Would we also like any feedback on the pages themselves now or wait until after this PR?

jestapinski (approved):
re-approved, thanks!

Dmytro-Shustra-EPAM:
Thanks for your comments!

PR #841, opened 2021-05-25
DEVHUB-457: Breadcrumbs improvements.

Dmytro-Shustra-EPAM:
This PR adds improvements for breadcrumbs and add CSS truncate effect for long text. [DEVHUB-457](https://jira.mongodb.org/browse/DEVHUB-457)

### DIFF
--- a/src/components/dev-hub/hero-banner.js
+++ b/src/components/dev-hub/hero-banner.js
@@ -52,11 +52,19 @@ const HeroBannerContainer = styled('div')`
     }
 `;
 
+const textTruncate = css`
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    max-width: 300px;
+`;
+
 const HeroBreadcrumb = styled(Breadcrumb)`
     line-height: 1;
     margin-bottom: ${HERO_BOTTOM_MARGIN};
     > a {
         font-size: ${fontSize.xsmall};
+        ${textTruncate}
     }
 `;
 
--- a/src/components/dev-hub/hero-banner.js
+++ b/src/components/dev-hub/hero-banner.js
@@ -53,10 +53,10 @@ const HeroBannerContainer = styled('div')`
 `;
 
 const textTruncate = css`
-    white-space: nowrap;
+    max-width: 300px;
     overflow: hidden;
     text-overflow: ellipsis;
-    max-width: 300px;
+    white-space: nowrap;
 `;
 
 const HeroBreadcrumb = styled(Breadcrumb)`
--- a/src/templates/article.js
+++ b/src/templates/article.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useMemo } from 'react';
 import styled from '@emotion/styled';
 import DocumentBody from '../components/DocumentBody';
 import ArticleShareFooter from '../components/dev-hub/article-share-footer';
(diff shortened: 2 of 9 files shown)


src/components/dev-hub/hero-banner.js
```diff
@@ -52,11 +52,19 @@ const HeroBannerContainer = styled('div')`
     }
 `;
 
+const textTruncate = css`
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    max-width: 300px;
```
jestapinski, line 59: Let's alphabetize this CSS please

src/templates/article.js
```diff
@@ -130,16 +135,21 @@ const Article = props => {
...
-        { label: 'Home', target: '/' },
-        { label: 'Learn', target: '/learn' },
-    ];
+    let breadcrumb = [...ARTICLE_BREADCRUMB];
     if (type && type.length) {
-        articleBreadcrumbs.push({
-            label: type[0].toUpperCase() + type.substring(1),
-            target: `/type/${getTagPageUriComponent(type)}`,
-        });
+        breadcrumb = [
```
jestapinski, line 140: Perhaps we could use the `splice` command https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice

This would let us initialize breadcrumb and then insert type at index 2 while deleting 0, what do you think?

jestapinski: Also, perhaps we could move the const into the component to avoid the redeclaration

src/templates/project.js
```diff
@@ -89,6 +95,11 @@ const Project = props => {
     const childNodes = getContent(dlv(content, 'children', []));
     const { siteUrl } = useSiteMetadata();
     const projectUrl = `${siteUrl}${props.pageContext.slug}`;
+    const breadcrumb = [
```
jestapinski, line 98: Similar thought, should we move the STUDENT_SPOTLIGHT_BREADCRUMBS into the component to avoid the redeclaration?

jestapinski (commented):
Looks nice so far! Can we check the breadcrumbs for `/academia/students/` as well? I believe that one is missing an end

src/templates/article.js
Dmytro-Shustra-EPAM: Great idea, Also I will add `memo` for this logic.

jestapinski:
Not sure if this was there before but we are missing a slash on the last breadcrumb? On your branch I see

```
{"@context":"https://schema.org","@type":"BreadcrumbList","itemListElement":[{"@type":"ListItem","position":1,"name":"Home","item":"https://developer.mongodb.com"},{"@type":"ListItem","position":2,"name":"Learn","item":"https://developer.mongodb.com/learn/"},{"@type":"ListItem","position":3,"name":"Quickstart","item":"https://developer.mongodb.com/type/quickstart/"},{"@type":"ListItem","position":4,"name":"Multi-Document ACID Transactions in MongoDB with Go","item":"https://developer.mongodb.comquickstart/golang-multi-document-acid-transactions/"}]}
```

For `quickstart/golang-multi-document-acid-transactions/`. Can we double check this please?

src/templates/article.js
```diff
@@ -125,21 +124,27 @@ const Article = props => {
         },
         ...rest
     } = props;
-    const slugWithAllSlashes = addLeadingSlashIfMissing(
-        addTrailingSlashIfMissing(slug)
-    );
-    const meta = { authors, slug: slugWithAllSlashes, title };
```
jestapinski, line 131: Did we mean to make this change? We want the leading and trailing slash for ratings but that seems to have been removed?

Dmytro-Shustra-EPAM: We just used slug on 149 line. And slugWithAllSlashes was unused.

jestapinski: We check for `slug` from `meta` [in this line](https://github.com/mongodb/devhub/blob/staging/src/components/dev-hub/feedback/feedback-container.js#L91) for the feedback. I didn't see if we added `slug` to `meta` anywhere else in the component?

Dmytro-Shustra-EPAM: Oh, I see, We use this logic for feedback modal. Thanks @jestapinski!

jestapinski (approved):
Great job, Thanks!

PR #859, opened 2021-06-09
Add Google Lighthouse Tooling into Github Actions CI

jestapinski:
This PR adds a change to our GitHub actions CI testing to add testing two URLs with Google Lighthouse. [This plugin](https://github.com/treosh/lighthouse-ci-action) allows us to identify web performance metrics which can be used to judge the overall health of the site and prevent significant regression.

Google Lighthouse, if you are unfamiliar, is a tool which monitors overall site health and performance in specific areas. It also helps with Search Engine rankings (kind of). You can read up on it [here](https://developers.google.com/web/tools/lighthouse)!


jestapinski:
Let's try it and if something happens we can revert

PR #861, opened 2021-06-13
Devhub 606 learn tabs/leafy tabs

harikakotipalli:
Creating new component Learn tabs wrapping LeafyGreen tabs underneath.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "@leafygreen-ui/icon": "^9.0.0",
     "@leafygreen-ui/icon-button": "^9.1.2",
     "@leafygreen-ui/leafygreen-provider": "^2.0.2",
-    "@leafygreen-ui/tabs": "^3.0.1",
+    "@leafygreen-ui/tabs": "^5.1.2",
     "@leafygreen-ui/text-input": "^5.0.10",
     "@reach/router": "^1.3.4",
     "@testing-library/react-hooks": "^5.1.1",
--- a/src/components/dev-hub/learn-tabs.js
+++ b/src/components/dev-hub/learn-tabs.js
@@ -3,8 +3,6 @@ import { css } from '@emotion/react';
 import styled from '@emotion/styled';
 import { Tabs as LeafyTabs, Tab } from '@leafygreen-ui/tabs';
 import Button from './button';
-import { TabContext } from './tab-context';
-import { screenSize } from './theme';
 
 const TabButtonStyle = () => css`
     border-radius: 0;
@@ -23,9 +21,9 @@ const StyledTabButton = styled(Button)`
 `;
 
 const TabsStyle = () => css`
-    @media ${screenSize.xlargeAndUp} {
+    @media (min-width: 1048px) {
         button:first-child {
-            margin-right: 74%;
+            margin-right: 73%;
         }
     }
     button[aria-selected='true'] {
@@ -40,39 +38,25 @@ const StyledTabs = styled(LeafyTabs)`
     ${TabsStyle}
 `;
 
-// Name anonymous tabsets by alphabetizing their tabids and concatenating with a forward slash
-const generateAnonymousTabsetName = tabIds => [...tabIds].sort().join('/');
-
 const TabButton = ({ ...props }) => (
     <StyledTabButton renderAsButton {...props} />
 );
 
 const Tabs = ({ tabsList, handleClick, activeItem }) => {
-    const { activeTabs, setActiveTab } = useContext(TabContext);
-    const tabsetName = generateAnonymousTabsetName(tabsList);
     const [activeTabIndex, setActiveTabIndex] = useState(0);
+    const adjustTabIndex = activeTab => {
+        const index = tabsList.indexOf(activeTab);
+        if (index !== -1) {
+            setActiveTabIndex(index);
+        }
+    };
 
     /*
     to set the value from tab context in nav bar
     */
     useEffect(() => {
-        setActiveTab({
-            tabset: tabsetName,
-            value: activeItem,
-        });
+        adjustTabIndex(activeItem);
     }, [activeItem]);
-
-    /*
-    to set the active tab index if
-    1. setActiveTab triggered from context changes in the nav bar
-    2. setActiveTab triggered from using call back on click
-    */
-    useEffect(() => {
-        const index = tabsList.indexOf(activeTabs[tabsetName]);
-        if (index !== -1) {
-            setActiveTabIndex(index);
-        }
-    }, [activeTabs, tabsList]);
 
     /*
     to set the active item upon user click on any leafy tabs
@@ -80,10 +64,7 @@ const Tabs = ({ tabsList, handleClick, activeItem }) => {
     const onClick = useCallback(
         index => {
             const tabValue = tabsList[index];
-            setActiveTab({
-                tabset: tabsetName,
-                value: tabValue,
-            });
+            adjustTabIndex(tabValue);
             handleClick(tabValue);
         },
         [tabsList]
--- a/src/components/dev-hub/learn-tabs.js
+++ b/src/components/dev-hub/learn-tabs.js
@@ -16,10 +16,9 @@ const StyledTabButton = styled(Button)`
 `;
 
 const StyledTabs = styled(LeafyTabs)`
+    margin: 0 32px;
     justify-items: flex-end;
-
     button {
-        margin: 5px;
         :first-of-type {
             margin-right: auto;
         }
--- a/src/components/dev-hub/learn-tabs.js
+++ b/src/components/dev-hub/learn-tabs.js
@@ -2,6 +2,7 @@ import React, { useCallback, useEffect, useState } from 'react';
 import styled from '@emotion/styled';
 import { Tabs as LeafyTabs, Tab } from '@leafygreen-ui/tabs';
 import Button from './button';
+import { size } from './theme';
 
 const StyledTabButton = styled(Button)`
     border-radius: 0;
@@ -16,8 +17,9 @@ const StyledTabButton = styled(Button)`
 `;
 
 const StyledTabs = styled(LeafyTabs)`
-    margin: 0 32px;
+    margin: 0 ${size.large};
     justify-items: flex-end;
+
     button {
         :first-of-type {
             margin-right: auto;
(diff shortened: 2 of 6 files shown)


src/components/dev-hub/learn-tabs.js
```diff
@@ -0,0 +1,98 @@
...
+        color: ${({ theme }) => theme.colorMap.darkGreen};
+        :after {
+            background-color: ${({ theme }) => theme.colorMap.darkGreen};
+        }
+    }
+`;
+
+const StyledTabs = styled(LeafyTabs)`
+    ${TabsStyle}
+`;
```
jestapinski, line 39:
```suggestion
const TabsStyle = (theme) => css`
    @media (min-width: 1048px) {
        button:first-child {
            margin-right: 73%;
        }
    }
    button[aria-selected='true'] {
        color: ${theme.colorMap.darkGreen};
        :after {
            background-color: ${theme.colorMap.darkGreen};
        }
    }
`;

const StyledTabs = styled(LeafyTabs)`
    ${({theme}) => TabsStyle(theme)};
`;
```

`css` does not use `theme` in the same way that `styled` does, so we explicitly have to pass it if we want to use `css`. Another thought is could we just put all of the style into `StyledTabs` like the below?

```
const StyledTabs = styled(LeafyTabs)`
        @media (min-width: 1048px) {
        button:first-child {
            margin-right: 73%;
        }
    }
    button[aria-selected='true'] {
        color: ${({ theme }) => theme.colorMap.darkGreen};
        :after {
            background-color: ${({ theme }) => theme.colorMap.darkGreen};
        }
    }
`;
```

That will work since `theme` is always passed to `styled`

src/components/dev-hub/learn-tabs.js
```diff
@@ -0,0 +1,98 @@
...
+        color: ${({ theme }) => theme.colorMap.darkGreen};
+        :after {
+            background-color: ${({ theme }) => theme.colorMap.darkGreen};
+        }
+    }
+`;
+
+const StyledTabButton = styled(Button)`
+    ${TabButtonStyle}
+`;
```
jestapinski, line 21:
```suggestion
const StyledTabButton = styled(Button)`
    border-radius: 0;
    :active,
    :hover,
    :focus {
        color: ${({ theme }) => theme.colorMap.darkGreen};
        :after {
            background-color: ${({ theme }) => theme.colorMap.darkGreen};
        }
    }
`;
```

Similar thought here, could we put the styles directly into `styled`? Then `theme` will work as we expect here since it is always passed to `styled` and not `css`

src/components/dev-hub/learn-tabs.js
```diff
@@ -0,0 +1,98 @@
...
+
+const TabsStyle = () => css`
+    @media (min-width: 1048px) {
```
jestapinski, line 24: Could we use the media query `largeAndUp` from the `theme.js` file? It is close enough and would help us standardize layout changes?

src/components/dev-hub/learn-tabs.js
```diff
@@ -0,0 +1,98 @@
...
+        button:first-child {
+            margin-right: 73%;
```
jestapinski, line 26: Curious where does this 73% come from? Was only doing this above 1023px Mike's idea?

src/components/dev-hub/learn-tabs.js
```diff
@@ -0,0 +1,98 @@
...
+        }
+    }
+    button[aria-selected='true'] {
+        color: ${({ theme }) => theme.colorMap.darkGreen};
```
jestapinski, line 30: Nice use of the `theme` here!

src/components/dev-hub/learn-tabs.js
```diff
@@ -0,0 +1,98 @@
...
+                selected={activeTabIndex}
+                setSelected={onClick}
+            >
+                {tabsList.map((tab, index) => {
+                    const tabId = tab;
+                    return (
+                        <Tab
+                            data-tabid={tabId}
+                            role="tab"
+                            key={index}
```
jestapinski, line 88:
```suggestion
                            key={tab}
```

Generally for the key in React it actually isn't good practice to use `index`! In this case it is likely not going to lead to issues, [however it is an anti-pattern we should avoid](https://robinpokorny.medium.com/index-as-a-key-is-an-anti-pattern-e0349aece318).

src/components/dev-hub/learn-tabs.js
```diff
@@ -0,0 +1,98 @@
...
+        <>
+            <StyledTabs
+                as={TabButton}
+                darkMode
+                data-test="tabs"
+                selected={activeTabIndex}
+                setSelected={onClick}
+            >
+                {tabsList.map((tab, index) => {
+                    const tabId = tab;
```
jestapinski, line 83: Do we need this reassignment or can we use `tab` below instead?

src/components/dev-hub/learn-tabs.js
```diff
@@ -0,0 +1,98 @@
...
+    to set the value from tab context in nav bar
+    */
+    useEffect(() => {
+        adjustTabIndex(activeItem);
+    }, [activeItem]);
+
+    /*
+    to set the active item upon user click on any leafy tabs
+    */
+    const onClick = useCallback(
```
jestapinski, line 64: Nice! This is great

src/components/dev-hub/learn-tabs.js
```diff
@@ -0,0 +1,98 @@
...
+    }
+    button[aria-selected='true'] {
+        color: ${({ theme }) => theme.colorMap.darkGreen};
+        :after {
+            background-color: ${({ theme }) => theme.colorMap.darkGreen};
+        }
+    }
+`;
+
+const StyledTabs = styled(LeafyTabs)`
```
jestapinski, line 37: One other thought here is for mobile we could stack if we wanted by adding `flex-direction: column`

jestapinski (commented):
Added some initial comments! Looking great so far.

Specifically, my thoughts around styled vs css will help with that `mount` test error you were seeing.

src/components/dev-hub/learn-tabs.js
harikakotipalli: Mike said he doesn't like the stacking approach we currently have. He prefers a drop down( which is his ideal choice) , if not he is okay with them lining next to each other.

src/components/dev-hub/learn-tabs.js
harikakotipalli: 73% is the maximum margin and 1048 is the screen size ,I could go without squishing the tab contents.Definitely want your suggestions on this.

src/components/dev-hub/learn-tabs.js
harikakotipalli: I like this approach. Will make changes.

src/components/dev-hub/learn-tabs.js
jestapinski: Oh okay I see, let's start with them next to each other as we have here and then next step would be a dropdown

src/components/dev-hub/learn-tabs.js
jestapinski: What about something like this in `TabsStyle` css?:

```
justify-items: flex-end;
@media (min-width: 1048px) {
    button:first-of-type {
        margin-right: auto;
    }
}
```

harikakotipalli: I don't need the min width any more. It work's better without the min-width.

src/components/dev-hub/learn-tabs.js
```diff
@@ -0,0 +1,83 @@
...
+        index => {
+            const tabValue = tabsList[index];
+            adjustTabIndex(tabValue);
+            handleClick(tabValue);
+        },
+        [tabsList]
+    );
+
+    return (
+        <>
```
jestapinski, line 65: Last nitpick from me, we shouldn't need the React fragments `<></>` because we are rendering one component here

jestapinski (approved):
Looks awesome! I had a last nitpick but this PR is good to go

jestapinski:
Just saw this actually when checking responsiveness, do we want to add some slight `margin` to the tabs so they do not run against the side of the device?

Screen Shot 2021-06-16 at 1 13 22 PM

src/components/dev-hub/learn-tabs.js
harikakotipalli: Good catch and thanks for the clarification.I had this question, but the other tabs.js had it, and I believe even that is a single component, so I just left it as is thinking may be it is our format. Will fix it.

jestapinski: Oh good catch, I will look into it. That may just be an error. So long as we return one component then it is fine to omit.

src/components/dev-hub/learn-tabs.js
```diff
@@ -0,0 +1,83 @@
...
+            background-color: ${({ theme }) => theme.colorMap.darkGreen};
+        }
+    }
+`;
+
+const StyledTabs = styled(LeafyTabs)`
+    justify-items: flex-end;
+
+    button {
+        margin: 5px;
```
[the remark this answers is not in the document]
harikakotipalli: oops , fixed it. Sorry for going back and forth on CSS. Trust me will make it easy for your next pull reqs.

src/components/dev-hub/learn-tabs.js
```diff
@@ -0,0 +1,82 @@
...
+    :focus {
+        color: ${({ theme }) => theme.colorMap.darkGreen};
+        :after {
+            background-color: ${({ theme }) => theme.colorMap.darkGreen};
+        }
+    }
+`;
+
+const StyledTabs = styled(LeafyTabs)`
+    margin: 0 32px;
```
jestapinski, line 19: Awesome! Last nitpick is can we use the `large` size option from `src/components/dev-hub/theme.js` in here? [Here is is on GitHub](https://github.com/mongodb/devhub/blob/staging/src/components/dev-hub/theme.js#L47)

No worries! definitely will take a bit of adjustment but for a first PR this is fantastic so please recognize that! Personally, I get very nitpicky with the PRs

jestapinski (approved):
re-approved, thanks!

package.json
```diff
@@ -37,7 +37,7 @@
     "@leafygreen-ui/icon": "^9.0.0",
     "@leafygreen-ui/icon-button": "^9.1.2",
     "@leafygreen-ui/leafygreen-provider": "^2.0.2",
-    "@leafygreen-ui/tabs": "^3.0.1",
+    "@leafygreen-ui/tabs": "^5.1.2",
```
jestapinski, line 40: For the version change we should install using `npm i @leafygreen-ui/tabs@5.1.2`, which updates the lockfile as well for us since that is also a record of all of our dependencies.

jestapinski: Oh whoops just saw that change, please disregard.

PR #863, opened 2021-06-14
Let inline Strapi elements be inline

jestapinski:
[This PR for an Article](https://docs-mongodbcom-staging.corp.mongodb.com/CI/devhub/jordanstapinski/fix-inline-elements/how-to/realm-data-types/)
[Existing Article](http://docs-devhub-staging.s3-website-us-east-1.amazonaws.com/25c77c5/devhub/docsworker-xlarge/strapi/how-to/realm-data-types/)

This PR checks if someone in Strapi uses HTML that is just a line break. We only want to use `span` for this to be inline and not a `div` since that would make it a block with height. The ask is to remove breaks from `div` and make them a span.

Check out this PR for an article and that same article as it is now from Strapi to see the difference!

### DIFF
--- a/src/components/dev-hub/raw-html.js
+++ b/src/components/dev-hub/raw-html.js
@@ -1,12 +1,17 @@
 import React from 'react';
 import sanitizeHtml from 'sanitize-html';
 
+const isInlineElement = html => html === '<br />';
+
 const RawHTML = ({ nodeData }) => {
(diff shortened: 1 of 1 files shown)


jestapinski:
I'm going to merge this now since a stakeholder is depending on it for their work today. If we have questions on it I am happy to reopen and change

PR #870, opened 2021-06-17
Add new ENV option for username customization for staging links

jestapinski:
This PR adds support for a new environment variable which we can use which allows us to separate which username we use to get reStructuredText content from which username we use for a staging link. This allows us to properly build and stage the platform.

To test:
- Open .env.production
- Add a new line `GATSBY_STAGING_USER=<your computer username, such as sophia.li>`
- run `npm run clean; npm run build; make stage`
- The site should build and stage, with all images/fonts looking good


jestapinski:
🥳

PR #871, opened 2021-06-21
DEVHUB-482: Add Cypress UI Testing for Student Spotlight Section

jestapinski:
No staging link

This PR adds some testing for one portion of our site under `/academia/`. The main goal is to test our integration with Strapi, to ensure non-static content fields are populated appropriately. Mainly this includes:
- Student projects we render
- Featured student projects
- Grids of student projects
- General navigation

As long as tests pass, this should be all good


src/queries/projects.js
```diff
@@ -13,6 +13,7 @@ query Projects {
     allStrapiProjects {
       nodes {
```
harikakotipalli, line 14: is this name different from bio name

jestapinski: Yeah it is, I was noticing we were using `bio name` which is not available anymore in Strapi (we moved it to just `name`). The test does look for a student's name so this is covered there.

If needed I am happy to make a staging link to compare w prod

jestapinski:
Going to merge, but happy to discuss if there are any concerns. Thanks!

PR #875, opened 2021-06-23
DEVHUB-710: Use withPrefix for Search Images

jestapinski:
[Staging Search on Learn Page](https://docs-mongodbcom-staging.corp.mongodb.com/master/devhub/jordanstapinski/DEVHUB-710/learn/?content=Articles&text=world)
[Existing new domain search](https://www.mongodb.com/developer/learn/?content=Articles&text=world#main)

This PR fixes a bug where internal images from a search were not being applied with the `pathPrefix`, meaning instead of loading an image from `/developer/images/img` we were looking at `/images/img` which fails.

The solution was to see if an image file is local (starts with a slash) and if so, apply Gatsby's `withPrefix` function to tack on either `/developer` for the new domain or something else for a different build.

To verify, check out the staging link and see the images load whereas on .com/developer they do not!

### DIFF
--- a/gatsby-config.prod-new.js
+++ b/gatsby-config.prod-new.js
@@ -10,7 +10,7 @@ require('dotenv').config({
 const metadata = getMetadata();
 
 module.exports = {
-    pathPrefix: 'developer',
+    pathPrefix: '/developer',
     plugins: [
         'gatsby-plugin-react-helmet',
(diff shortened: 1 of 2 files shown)


sophiali23 (approved):
nice!

jestapinski:
🥳  Thanks!

PR #917, opened 2021-07-28
DEVHUB-761: Application Form - Managing State of UI

sophiali23:
## Links:

-   [JIRA Ticket](https://jira.mongodb.org/browse/DEVHUB-761)
-   [Staging Link](https://docs-mongodbcom-staging.corp.mongodb.com/CI/devhub/sophia.li/DEVHUB-761/community-champions/)

## Description:

-   This PR handles form submissions and calls the `submitCommunityChampionApplication` Realm function which inserts application responses to the `devhub.community_champion_application_responses` collection. 

## Testing

Success case:
- Fill out all the fields of the form with valid input, then submit and verify that the "Thank you for applying!" success modal shows up right after. Then, verify that your response was properly inserted into our [collection](https://cloud.mongodb.com/v2/5eb9a4925f91995d7c0e426a#metrics/replicaSet/5ec43e18e19df87d804548cf/explorer/devhub/community_champion_application_responses/find)

Error case:
- To test this, I opened the form then turned off my wifi. Fill out all the fields of the form with valid input, then click "Submit". Verify that a red "Your submission failed. Please try again." message shows up at the top of the form. Also verify that no new documents have been inserted into the  [collection](https://cloud.mongodb.com/v2/5eb9a4925f91995d7c0e426a#metrics/replicaSet/5ec43e18e19df87d804548cf/explorer/devhub/community_champion_application_responses/find)

-   [ ] Documentation Updates
-   [ ] Product Review
-   [x] Design Review

### DIFF
--- a/src/components/dev-hub/academia-sign-up-form.js
+++ b/src/components/dev-hub/academia-sign-up-form.js
@@ -66,15 +66,6 @@ const InstitutionSection = styled('div')`
     }
 `;
 
-const callRealm = async (data, callback) => {
-    try {
-        const res = await submitAcademiaForm(data);
-        res && callback(true);
-    } catch {
-        callback(false);
-    }
-};
-
 const AcademiaSignUpForm = React.memo(({ setSuccess, success, ...props }) => {
     const [firstName, setFirstName] = useState('');
     const [lastName, setLastName] = useState('');
@@ -101,11 +92,10 @@ const AcademiaSignUpForm = React.memo(({ setSuccess, success, ...props }) => {
             institution_type: institutionType,
             agree_to_email: agreeToEmail,
         };
-        const callback = hasSuccess => {
-            setCanSubmit(!hasSuccess);
-            setSuccess(hasSuccess);
-        };
-        callRealm(data, callback);
+        const response = await submitAcademiaForm(data);
+
+        setSuccess(response.success);
+        setCanSubmit(!response.success);
     };
 
     const onEmailInvalid = useCallback(
--- a/src/components/dev-hub/academia-sign-up-form.js
+++ b/src/components/dev-hub/academia-sign-up-form.js
@@ -66,6 +66,15 @@ const InstitutionSection = styled('div')`
     }
 `;
 
+const callRealm = async (data, callback) => {
+    try {
+        const res = await submitAcademiaForm(data);
+        res && callback(true);
+    } catch {
+        callback(false);
+    }
+};
+
 const AcademiaSignUpForm = React.memo(({ setSuccess, success, ...props }) => {
     const [firstName, setFirstName] = useState('');
     const [lastName, setLastName] = useState('');
@@ -92,10 +101,11 @@ const AcademiaSignUpForm = React.memo(({ setSuccess, success, ...props }) => {
             institution_type: institutionType,
             agree_to_email: agreeToEmail,
         };
-        const response = await submitAcademiaForm(data);
-
-        setSuccess(response.success);
-        setCanSubmit(!response.success);
+        const callback = hasSuccess => {
+            setSuccess(hasSuccess);
+            setCanSubmit(!hasSuccess);
+        };
+        callRealm(data, callback);
     };
 
     const onEmailInvalid = useCallback(
--- a/src/components/dev-hub/academia-sign-up-form.js
+++ b/src/components/dev-hub/academia-sign-up-form.js
@@ -102,8 +102,8 @@ const AcademiaSignUpForm = React.memo(({ setSuccess, success, ...props }) => {
             agree_to_email: agreeToEmail,
         };
         const callback = hasSuccess => {
-            setSuccess(hasSuccess);
             setCanSubmit(!hasSuccess);
+            setSuccess(hasSuccess);
         };
         callRealm(data, callback);
     };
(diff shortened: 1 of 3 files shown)


src/components/dev-hub/community-champion-form.js
```diff
@@ -175,7 +199,11 @@ const Form = React.memo(({ setSuccess, success }) => {
                 variant="default"
             >
                 {experienceOptions.map((option, index) => (
-                    <Radio key={index} value={option}>
+                    <Radio
+                        checked={experience == option}
```
sophiali23, line 203: adding `checked` attribute resolves this warning I was having when switching between radio buttons: 

Screen Shot 2021-07-28 at 12 15 09 PM

src/components/dev-hub/community-champion-form.js
```diff
@@ -98,7 +108,21 @@ const Form = React.memo(({ setSuccess, success }) => {
...
+            first_name: firstName,
+            last_name: lastName,
+            email,
+            experience,
+            bio,
+            agree_to_email: agreeToEmail,
+        };
+        const callback = hasSuccess => {
+            setCanSubmit(!hasSuccess);
+            setSuccess(hasSuccess);
```
sophiali23, line 123: am noticing that the order of these two make a difference, calling setSuccess before setCanSubmit results in the following warning: 

Screen Shot 2021-07-28 at 11 05 47 AM

I believe this is because updating success state causes the modal to re-render and unmounts the form to mount the "Thanks for applying!" success component which results in setCanSubmit being a no-op since form has been unmounted (and canSubmit is one of the form's states). 

I wonder if we're also getting this warning on the project sign up form when we submit since setSuccess is also being called first https://github.com/mongodb/devhub/blob/staging/src/components/dev-hub/project-sign-up-form.js#L57-L58 ...haven't tested it yet since i didn't want to set off the trigger to send an email lol

jestapinski: Thanks for calling this out, I think your suggestion removes that possibility since if the component should re-render, only then do we update state. Let's go ahead and add that in! I'll add some more comments to the PR now :)

src/components/dev-hub/academia-sign-up-form.js
```diff
@@ -66,6 +66,15 @@ const InstitutionSection = styled('div')`
...
 `;
 
+const callRealm = async (data, callback) => {
+    try {
+        const res = await submitAcademiaForm(data);
+        res && callback(true);
+    } catch {
+        callback(false);
+    }
+};
```
jestapinski, line 76: I like this idea because we abstract the `try/catch` to its own function but something feels a bit off to me in the `callback` piece taking true/false.

I think what I was expecting was something more like the Javascript Promise API, where there are separate handlers for `onSuccess` and `onFailure`. Then we also can case on when, for example, to run `setCanSubmit` a bit better. I think, for example, imagine we did the following:

```
const callRealm = async (data, onSuccess, onFailure) => {
    try {
        const res = await submitAcademiaForm(data);
        res && onSuccess();
    } catch {
        onFailure();
    }
};
```

then later (with the suggestion to avoid code on an unmounted component), we can simply do this:

```
callRealm(data,() => setSuccess(true), () => {setSuccess(false); setCanSubmit(true)};
```

Why I feel this is a bit better is:
- We clearly separate what code runs on a successful submission vs a failure
- We then can pass more to the `onFailure`, such as a failure message without having to worry about how general `callback` would be

Honestly though I'm not too opinionated on this, but I did get a weird vibe when looking through this and I think the callback use is why (or could we update `submitAcademiaForm` directly to add the try/catch maybe and remove this logic from here?) Idk there are a couple options

src/components/dev-hub/community-champion-form.js
```diff
@@ -175,7 +199,11 @@ const Form = React.memo(({ setSuccess, success }) => {
                 variant="default"
             >
                 {experienceOptions.map((option, index) => (
-                    <Radio key={index} value={option}>
+                    <Radio
+                        checked={experience == option}
+                        key={index}
```
jestapinski, line 204:
```suggestion
                        key={option}
```

Using `index` for key is typically an anti-pattern in React actually so let's avoid any potential weird bugs by swapping to another unique value here please!

https://robinpokorny.com/blog/index-as-a-key-is-an-anti-pattern/

jestapinski (commented):
Nice! I had a couple opinions I'm curious what you think. Works well!

src/components/dev-hub/academia-sign-up-form.js
sophiali23: i do like the idea of separating the two handlers better than using an if statement in a general `callback`! i think it also makes the code easier to understand. and yeah, i dont see the point in keeping the logic there, i can update `submitAcademiaForm` to have the try/catch

src/components/dev-hub/community-champion-form.js
sophiali23: yess! forgot that was a thing haha

src/utils/devhub-api-stitch.js
```diff
@@ -74,6 +74,22 @@ export const submitStudentSpotlightProject = async projectData => {
     return result;
 };
 
+export const submitCommunityChampionApplication = async (
+    communityChampionData,
+    onSuccess,
+    onFailure
+) => {
+    try {
+        const result = await callDevhubAPIStitchFunction(
```
jestapinski, line 83: 👏  putting this logic in here is awesome since we abstract it from the form entirely, nice!

jestapinski (approved):
🥳

PR #921, opened 2021-07-30
Community Champions - set up profile page template

sophiali23:
## Links:

-   [JIRA Ticket](https://jira.mongodb.org/browse/DEVHUB-784)
-   [Staging Link](https://docs-mongodbcom-staging.corp.mongodb.com/CI/devhub/sophia.li/DEVHUB-784/community-champions/chris-dellaway/)

## Description:

-   This PR sets up the creation of Community Champion profile pages!

## Testing

-   Verify that there are pages at /community-champions/firstname-lastname for the champions shown at the bottom [here](https://docs-mongodbcom-staging.corp.mongodb.com/CI/devhub/sophia.li/DEVHUB-784/community-champions/). Profile pages currently only have breadcrumbs shown at the top.

-   [ ] Documentation Updates
-   [ ] Product Review
-   [ ] Design Review

### DIFF
--- a/src/queries/community-champions.js
+++ b/src/queries/community-champions.js
@@ -1,0 +1,30 @@
+export const communityChampions = `
+    query CommunityChampions {
+        allStrapiCommunityChampions(
+            sort: { fields: [firstName, middleName, lastName] }
+        ) {
+            nodes {
+                id
+                firstName
+                middleName
+                lastName
+                location
+                title
+                image {
+                    url
+                }
+                quote
+                languagesSpoken
+                bio
+                BlogsAndPublications {
+                  link
+                  title
+                }
+                Certifications {
+                  isDBAAssociateCertified
+                  isDeveloperAssociateCertified
+                }
+            }
+        }
+    }
+ `;
--- a/src/queries/community-champions.js
+++ b/src/queries/community-champions.js
@@ -1,10 +1,7 @@
 export const communityChampions = `
     query CommunityChampions {
-        allStrapiCommunityChampions(
-            sort: { fields: [firstName, middleName, lastName] }
-        ) {
+        allStrapiCommunityChampions {
             nodes {
-                id
                 firstName
                 middleName
                 lastName
--- a/src/templates/community-champion-profile.js
+++ b/src/templates/community-champion-profile.js
@@ -1,0 +1,55 @@
+import React from 'react';
+import styled from '@emotion/styled';
+import Breadcrumb from '~components/dev-hub/breadcrumb';
+import Layout from '~components/dev-hub/layout';
+import { fontSize, screenSize, size } from '~components/dev-hub/theme';
+
+const BREADCRUMB_BOTTOM_MARGIN = '40px';
+
+const ContentContainer = styled('div')`
+    padding: ${size.default} ${size.xxlarge};
+    @media ${screenSize.upToLarge} {
+        padding: ${size.mediumLarge} ${size.default};
+    }
+`;
+
+const StyledBreadcrumb = styled(Breadcrumb)`
+    line-height: 1;
+    margin-bottom: ${BREADCRUMB_BOTTOM_MARGIN};
+    > a {
+        font-size: ${fontSize.xsmall};
+    }
+    @media ${screenSize.upToLarge} {
+        margin-bottom: ${size.large};
+    }
+`;
+
+const CommunityChampionProfile = props => {
+    const {
+        pageContext: { champion, slug },
+    } = props;
+    const { firstName, middleName, lastName } = champion;
+    const fullName = [firstName, middleName, lastName].join(' ');
+    const championProfileBreadcrumbs = [
+        { label: 'Home', target: '/' },
+        {
+            label: 'MongoDB Community Champions',
+            target: '/community-champions',
+        },
+        {
+            label: `${fullName}`,
+            target: slug,
+        },
+    ];
+    return (
+        <Layout>
+            <ContentContainer>
+                <StyledBreadcrumb>
+                    {championProfileBreadcrumbs}
+                </StyledBreadcrumb>
+            </ContentContainer>
+        </Layout>
+    );
+};
+
+export default CommunityChampionProfile;
--- a/src/templates/community-champion-profile.js
+++ b/src/templates/community-champion-profile.js
@@ -6,6 +6,14 @@ import { fontSize, screenSize, size } from '~components/dev-hub/theme';
 
 const BREADCRUMB_BOTTOM_MARGIN = '40px';
 
+const CHAMPION_PROFILE_BREADCRUMBS_PREFIX = [
+    { label: 'Home', target: '/' },
+    {
+        label: 'MongoDB Community Champions',
+        target: '/community-champions',
+    },
+];
+
 const ContentContainer = styled('div')`
     padding: ${size.default} ${size.xxlarge};
     @media ${screenSize.upToLarge} {
@@ -30,17 +38,12 @@ const CommunityChampionProfile = props => {
     } = props;
     const { firstName, middleName, lastName } = champion;
     const fullName = [firstName, middleName, lastName].join(' ');
-    const championProfileBreadcrumbs = [
-        { label: 'Home', target: '/' },
-        {
-            label: 'MongoDB Community Champions',
-            target: '/community-champions',
-        },
+    const championProfileBreadcrumbs = CHAMPION_PROFILE_BREADCRUMBS_PREFIX.push(
         {
             label: `${fullName}`,
             target: slug,
-        },
-    ];
+        }
+    );
     return (
         <Layout>
             <ContentContainer>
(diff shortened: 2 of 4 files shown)


src/templates/community-champion-profile.js
```diff
@@ -0,0 +1,55 @@
...
+
+const ContentContainer = styled('div')`
+    padding: ${size.default} ${size.xxlarge};
+    @media ${screenSize.upToLarge} {
+        padding: ${size.mediumLarge} ${size.default};
+    }
+`;
+
+const StyledBreadcrumb = styled(Breadcrumb)`
+    line-height: 1;
```
jestapinski, line 17: This is clever, I've never seen it before. Nice!

src/utils/setup/create-community-champion-profile-pages.js
```diff
@@ -0,0 +1,34 @@
+import path from 'path';
+import { communityChampions } from '../../queries/community-champions';
+
+const createCommunityChampionProfilePage = async (champion, createPage) => {
+    const { firstName, lastName } = champion;
+    const slug = `/community-champions/${firstName.toLowerCase()}-${lastName.toLowerCase()}`;
```
jestapinski, line 6: Do any of the champions have any non-letter characters such as spaces or apostraphes in their names? For example d'arte has an `'` and would need to be escaped or such for a URL?

src/templates/community-champion-profile.js
```diff
@@ -0,0 +1,55 @@
...
+        margin-bottom: ${size.large};
+    }
+`;
+
+const CommunityChampionProfile = props => {
+    const {
+        pageContext: { champion, slug },
+    } = props;
+    const { firstName, middleName, lastName } = champion;
+    const fullName = [firstName, middleName, lastName].join(' ');
```
jestapinski, line 32: 👍

src/templates/community-champion-profile.js
```diff
@@ -0,0 +1,55 @@
...
+    @media ${screenSize.upToLarge} {
+        padding: ${size.mediumLarge} ${size.default};
+    }
+`;
+
+const StyledBreadcrumb = styled(Breadcrumb)`
+    line-height: 1;
+    margin-bottom: ${BREADCRUMB_BOTTOM_MARGIN};
+    > a {
+        font-size: ${fontSize.xsmall};
```
jestapinski, line 20: Just curious is this different from the rest of the breadcrumbs on the platform?

src/utils/setup/create-community-champion-profile-pages.js
```diff
@@ -0,0 +1,34 @@
...
+
+export const createCommunityChampionProfilePages = async (
+    createPage,
+    graphql
+) => {
+    const championList = await getChampionListFromGraphql(graphql);
+    championList.forEach(
+        async champion =>
+            await createCommunityChampionProfilePage(champion, createPage)
+    );
```
jestapinski, line 33: Nice use of `async/await`! However I don't see us using any `await` in `createCommunityChampionProfilePage` so we likely don't need to `asnyc/await` at all here. I did see the same mistake for academia projects so I can see where maybe that came from :)

src/queries/community-champions.js
```diff
@@ -0,0 +1,30 @@
+export const communityChampions = `
+    query CommunityChampions {
+        allStrapiCommunityChampions(
+            sort: { fields: [firstName, middleName, lastName] }
+        ) {
+            nodes {
+                id
```
jestapinski, line 7: Are we going to use this field later on?

src/queries/community-champions.js
```diff
@@ -0,0 +1,30 @@
+export const communityChampions = `
+    query CommunityChampions {
+        allStrapiCommunityChampions(
+            sort: { fields: [firstName, middleName, lastName] }
```
jestapinski, line 4: Nice! Do we plan on re-using this query later? If not, we probably don't need to sort if this is just for creating the pages

src/templates/community-champion-profile.js
```diff
@@ -0,0 +1,55 @@
...
+        { label: 'Home', target: '/' },
+        {
+            label: 'MongoDB Community Champions',
+            target: '/community-champions',
+        },
+        {
+            label: `${fullName}`,
+            target: slug,
+        },
+    ];
```
jestapinski, line 43: Nice! Perhaps one slight optimization we can even do is split this up so elements 0 and 1 live in a constant outside of the component and we just `push` 

```
        {
            label: `${fullName}`,
            target: slug,
        },
```

Onto the end. So altogether it might be something like:

```
const CHAMPION_BREADCRUMB_PREFIX = [
        { label: 'Home', target: '/' },
        {
            label: 'MongoDB Community Champions',
            target: '/community-champions',
        },
];

const CommunityChampionProfile = props => {
...
const championProfileBreadcrumbs = championProfileBreadcrumbs.concat([{
            label: `${fullName}`,
            target: slug,
        }])
}
```

Honestly though, this might not even be worth doing so lmk what do you think?

jestapinski (commented):
Excellent start working with the gatsby-node.esm.js file! I had some general thoughts

src/templates/community-champion-profile.js
sophiali23: hmmm im quite indifferent on this...i'll add it in!

src/queries/community-champions.js
sophiali23: whoopsies i copied and pasted my old query into here and forgot to take out the id and sorting when modifying it, thanks for catching that!

src/templates/community-champion-profile.js
sophiali23: no, the breadcrumbs in hero banners also use this font size

src/utils/setup/create-community-champion-profile-pages.js
sophiali23: thanks for bringing this up! yes there is the ö in Michael Höller. should i encode them by using `encodeURIComponent`?

src/templates/community-champion-profile.js
sophiali23: was not my idea, i took it from here 😛  https://github.com/mongodb/devhub/blob/staging/src/components/dev-hub/hero-banner.js#L67

src/utils/setup/create-community-champion-profile-pages.js
jestapinski: yes!

src/templates/community-champion-profile.js
jestapinski: Cool, I'm happy either way but just wanted to call it out

src/templates/community-champion-profile.js
jestapinski: Ah yes, Gregg had some tricks like that up his sleeve

src/templates/community-champion-profile.js
sophiali23: TY!

src/utils/setup/create-community-champion-profile-pages.js
sophiali23: this also got me thinking...what if we have two champions with the same name? i know this is extremely unlikely but would having the slug be something that's guaranteed to be unique like /community-champions/<champion_id> be better?

src/utils/setup/create-community-champion-profile-pages.js
sophiali23: are async functions useless if there are no `await`s in them?

src/utils/setup/create-community-champion-profile-pages.js
sophiali23: it's failing to create Michael's page when I use `encodeURIComponent`...seems like a known issue - https://github.com/gatsbyjs/gatsby/issues/16765 🤔

jestapinski: Would https://github.com/gatsbyjs/gatsby/issues/16765#issuecomment-632400169 help?

Two with the same name seems a bit improbable tbh so not sure we need to worry about that. Also design/product has asked for the URL to contain the name for ease-of-sharing purposes

sophiali23: > Would [gatsbyjs/gatsby#16765 (comment)](https://github.com/gatsbyjs/gatsby/issues/16765#issuecomment-632400169) help?

page does get created if i use `decodeURIComponent` although im not sure what the point of decoding the uri component is if it was not encoded before? it just behaves the same as before without the encoding/decoding with the slug being /community-champions/michael-höller

> Two with the same name seems a bit improbable tbh so not sure we need to worry about that. Also design/product has asked for the URL to contain the name for ease-of-sharing purposes

ok! :)

src/utils/setup/create-community-champion-profile-pages.js
jestapinski: Yep, `asnyc` indicates that this function does something like a network call or so which should be handled properly, either by making the program `await` it or by using the JS Promise API. Even if nothing inside is actually asynchronous, it adds a bit of complexity when reasoning about the code, so if we don't need it we should avoid it!

src/utils/setup/create-community-champion-profile-pages.js
jestapinski: yeah hmm I'm also a bit confused here. If we don't need to decode then let's just not decode then my bad

jestapinski (approved):
👍

PR #935, opened 2021-08-05
Don't make community forum links internal

jestapinski:
## Links:

-   No JIRA Ticket
-   [Staging Link article fixed](https://docs-mongodbcom-staging.corp.mongodb.com/master/devhub/jordanstapinski/fix-old-community-url/article/introduction-to-modern-databases-mongodb-academia/?tck=feathome)
-   [Same article on prod with issue](https://www.mongodb.com/developer/article/introduction-to-modern-databases-mongodb-academia/?tck=feathome)

## Description:

-   This PR fixes a bug related to the subdomain swap in which we incorrectly parsed old links to the forums and made links internal. We should have not changed these links and leave the href as it was. The bug was that the `OLD_SUBDOMAIN_FORUMS_URL` pointed to `https://www.mongodb.com/developer/community/forums` and not `https://developer.mongodb.com/community/forums`

## Testing

-   I updated our unit test to be a bit stronger and check for absolute domains instead of assuming consts were correct being imported, which they were not.
-   I put an article with the issue as well as a staging link of the same article above. At the end both have a link to the forums, but only in the fixed one does it not 404

-   [ ] Documentation Updates
-   [ ] Product Review
-   [ ] Design Review

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -36,5 +36,5 @@ export const FEEDBACK_APP_ID = 'feedback-ihion';
 
 export const OLD_SITE_URL = 'https://developer.mongodb.com';
 export const SITE_URL = 'https://www.mongodb.com/developer';
-export const OLD_SUBDOMAIN_FORUMS_URL = `${SITE_URL}/community/forums/`;
+export const OLD_SUBDOMAIN_FORUMS_URL = `${OLD_SITE_URL}/community/forums/`;
 export const FORUMS_URL = `https://www.mongodb.com/community/forums/`;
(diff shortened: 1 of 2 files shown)


src/constants.js
```diff
@@ -36,5 +36,5 @@ export const FEEDBACK_APP_ID = 'feedback-ihion';
 
 export const OLD_SITE_URL = 'https://developer.mongodb.com';
 export const SITE_URL = 'https://www.mongodb.com/developer';
-export const OLD_SUBDOMAIN_FORUMS_URL = `${SITE_URL}/community/forums/`;
+export const OLD_SUBDOMAIN_FORUMS_URL = `${OLD_SITE_URL}/community/forums/`;
```
jestapinski, line 39: This was the culprit, we didn't swap this when flipping domains

PR #978, opened 2021-09-20
Devhub 820 academia form changes

harikakotipalli:
## Links:

-   [JIRA](https://jira.mongodb.org/browse/DEVHUB-820)
-   [Staging](http://developer-hub-staging.s3-website-us-east-1.amazonaws.com/CI/devhub/harika.kotipalli/DEVHUB-820-AcademiaFormChanges/)

## Description:

-   Academia for Educators changes as per Jira

## Testing

-   How might we test that this PR works as expected?
    Make sure all the changes requested are captured in the new form
-   Are there any automated tests related to this PR?

-   [ ] Documentation Updates
-   [ ] Product Review
-   [ ] Design Review : May be.


src/components/dev-hub/academia-sign-up-form.js
```diff
@@ -189,8 +264,30 @@ const AcademiaSignUpForm = React.memo(({ setSuccess, success, ...props }) => {
                     choices={institutionTypes}
                     onChange={e => setInstitutionType(e)}
                 />
+                <StyledSelect
```
jestapinski, line 267: There are a ton of options for countries, can we make the select scroll? Maybe set a max-height or something on it to allow `overflow` in css?

src/components/dev-hub/academia-sign-up-form.js
```diff
@@ -189,8 +264,30 @@ const AcademiaSignUpForm = React.memo(({ setSuccess, success, ...props }) => {
...
+                    narrow
```
jestapinski, line 268: Also `required`? Seems like everything should be required that is new but the course syllabus

jestapinski (commented):
Looking pretty good! Some questions on the form

We also will have to update the Realm data-access rules to allow these fields to be written to the DB and update the code which sends the email to them as well

src/components/dev-hub/academia-sign-up-form.js
harikakotipalli: Our current select doesn't support required, nor the leafy green ui.I raised a request to add required to leafy green select. For now I'm just erroring out for now, if the mandatory fields are not filled.

jestapinski: Got it, sounds good here!

src/components/dev-hub/academia-sign-up-form.js
harikakotipalli: Yeah, good advice. I will do that.

src/components/dev-hub/academia-sign-up-form.js
harikakotipalli: Unfortunately our select is built with a custom div and no html selects are used. It's not very easy to make the default fill out component, which we get for input/select tags out of box. https://jira.mongodb.org/browse/PD-1579- In case you are interested.

src/components/dev-hub/academia-sign-up-form.js
```diff
@@ -91,9 +136,18 @@ const AcademiaSignUpForm = React.memo(({ setSuccess, success, ...props }) => {
...
             institution_type: institutionType,
             agree_to_email: agreeToEmail,
+            instructor_type:
+                instructorType === 'other'
+                    ? instructorTypeDesc
+                    : instructorType,
+            instructor_interests: instructorInterest,
+            country: country,
+            course_name: courseName,
+            course_syllabus: courseSyllabus,
```
jestapinski, line 146: I think we are good to merge this so long as in Realm we add these to `onAcademiaRegistration` where the email is sent to the Academia team

jestapinski (approved):

harikakotipalli:
Update the realm function with new fields yes, but I don’t think it is
sending email, I mean I dint find that functionality in the realm function.

jestapinski:
Ah that is done via a DB trigger. After we insert the record there is a
trigger to call `onAcademiaRegistration` which sends the email via AWS SES.

